PR #177, opened 2013-11-24
moved tutorial features into features folder

fgdrf:
its a minor cleanup to keep features separated from plugins


fgdrf:
We should not refactor tutorials prior 1.5.0 release, therefor I close it

PR #185, opened 2015-02-09
[UDIG-1987] added optional oracle jdbc imports

fgdrf:
to resolve oracle.jdbc dependencies from another bundle then org.locationtech.udig.libs.oracle

* its relevant for Integrators to choose a pre-packaged bundle with their prefereded jdbc driver version and 
* have the posibility too use Utility classes from e.g. gt-jdbc-oracle (that would not work without optional imports)


fgdrf:
https://jira.codehaus.org/browse/UDIG-1987

PR #186, opened 2015-02-09
[UDIG-1897] avoid cast exception guessing a CRS

fgdrf:
https://jira.codehaus.org/browse/UDIG-1897

### DIFF
--- a/plugins/org.locationtech.udig.ui/src/org/locationtech/udig/ui/CRSChooser.java
+++ b/plugins/org.locationtech.udig.ui/src/org/locationtech/udig/ui/CRSChooser.java
@@ -22,7 +22,6 @@ import java.util.regex.Pattern;
 
 import org.locationtech.udig.internal.ui.UiPlugin;
 import org.locationtech.udig.ui.internal.Messages;
-
 import org.eclipse.core.runtime.CoreException;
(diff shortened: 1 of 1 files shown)


plugins/org.locationtech.udig.ui/src/org/locationtech/udig/ui/CRSChooser.java
```diff
@@ -475,7 +475,10 @@ protected CoordinateReferenceSystem createCRS( String code ) {
         for( Object object : ReferencingFactoryFinder.getCRSAuthorityFactories(null) ) {
             CRSAuthorityFactory factory = (CRSAuthorityFactory) object;
             try {
-                return (CoordinateReferenceSystem) factory.createObject(code);
+                IdentifiedObject identifiedObject = factory.createObject(code);
+                if (identifiedObject instanceof CoordinateReferenceSystem) {
+                    return (CoordinateReferenceSystem) factory.createObject(code);
+                }
             } catch (FactoryException e2) {
```
jodygarnett: Not sure about a better way, I tend to use the CRS.decode( string ) method. I think the internals of that method would be a good way to see how these factories are used.

However this instance of sanity check seems fine...

PR #187, opened 2015-02-12
revert emf model rebuild changes for layers view

egouge:
To undo some changes made when the EMF model was regenerated.  These changes caused the layer view to not update correctly.  More info: https://locationtech.org/mhonarc/lists/udig-dev/msg24060.html

### DIFF
--- a/plugins/org.locationtech.udig.project.edit/src/org/locationtech/udig/project/internal/provider/ProjectItemProviderAdapterFactory.java
+++ b/plugins/org.locationtech.udig.project.edit/src/org/locationtech/udig/project/internal/provider/ProjectItemProviderAdapterFactory.java
@@ -105,21 +105,10 @@ public class ProjectItemProviderAdapterFactory extends ProjectAdapterFactory imp
      */
     @Override
     public Adapter createLayerAdapter() {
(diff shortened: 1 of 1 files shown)


plugins/org.locationtech.udig.project.edit/src/org/locationtech/udig/project/internal/provider/ProjectItemProviderAdapterFactory.java
```diff
@@ -105,21 +105,10 @@ public Adapter createContextModelAdapter() {
      */
     @Override
     public Adapter createLayerAdapter() {
-        if (layerItemProvider == null) {
-            layerItemProvider = new LayerItemProvider(this);
-        }
-
-        return layerItemProvider;
+        return new LayerItemProvider(this);
```
jodygarnett: This is an interesting change - I guess it makes a unique provider each time (rather than a single one). This is only necessary if the provider is now statefull - is that now the case?

egouge:
It seems that way.  When you have more then one map the provider references the first map when sending events.  I did not dig too far into the code to figure out why this was occurring.

PR #189, opened 2015-02-22
[UDIG-1737] added importToStyle to all advanced StyleEditors

fgdrf:
See https://jira.codehaus.org/browse/UDIG-1737 for details


fgdrf:
missed to sign the commit http://www.eclipse.org/legal/CoO.php

PR #194, opened 2015-06-12
Fixed service sorting in catalog plugin on acquire

HendrikPeilke:
This pull request fixes a mistake inside the catalog plugin making it choose the worst service out of the list of possible instead of the best. Also added a test.

PR #195, opened 2015-06-17
Catalog metric test services only reacting on service id

HendrikPeilke:
This fixes the test failures brought to uDig by pull request https://github.com/locationtech/udig-platform/pull/194 by checking for the right id inside createService and not createParams. Both tests run fine now.


fgdrf:
Pull-Request build https://hudson.locationtech.org/udig/job/uDig-PR-NG/3/ indicated that the test is fixed. Thanks for the patch Hendrik

PR #199, opened 2015-10-06
fixed number formatting in CursorPosition tool

AlexanderRice:
This change takes care of a problem with the coordinate display on the Workbench status line. Formerly, in cases of negative longitudes of more than three integer digits, only the integer part of the number would be displayed, and the fractional part would be lopped off.

The cause was a test that failed to account for a minus sign. It would lop of the fractional part of the number if there were more than three characters before the decimal point. This would accommodate positive longitudes of three integer digits, but negative longitudes of only two integer digits (a problem for us North Americans).

I have left in the test that was causing the issue, but modified it to account for the minus sign.

I have attached before/after pictures in case it is not clear what I'm talking about.

before
after

### DIFF
--- a/plugins/org.locationtech.udig.tool.default/src/org/locationtech/udig/tools/internal/CursorPosition.java
+++ b/plugins/org.locationtech.udig.tool.default/src/org/locationtech/udig/tools/internal/CursorPosition.java
@@ -157,7 +157,8 @@ public class CursorPosition extends AbstractTool {
             String string = format.format(value);
 
             String[] parts = string.split("\\.");
-            if(parts[0].length()>3){
+            int prefix = parts[0].startsWith("-") ? 1 : 0;
+            if(parts[0].length()>3+prefix){
             	string = parts[0];
             }
             return string;


fgdrf:
IMO the fix looks good.

Since the ip-check fails, please have a look at CONTRIBUTING.textile and sign your commit to proceed.

fgdrf:
Just started a pull request build https://hudson.locationtech.org/udig/job/uDig-PR-NG/8/

fgdrf:
@AlexanderRice I'd like to merge your commit. Could you please sign your commit, maybe the latest post helps : https://t.co/HhoLbRjuKy . I your case, because you committed on your master branch, reset it to the last commit before relevant for this pull and commit changes with -s option. You could also create a branch (checkout -b) if file CursorPosition.java is marked as changed. HTH

AlexanderRice:
Hi Frank,

Sorry for not signing originally. Just wanted to acknowledge receipt of your email and will take a look when I get some free time.

Sincerely,
Alexander Rice

fgdrf:
@AlexanderRice Any news here? How can I help you finishing this pull request?

fgdrf:
@AlexanderRice Since this is only one commit I'd like to suggest to reset your branch 
`git reset HEAD~1`
re-commit with option `-s`
and push your branch again with `-f` option. This should help

fgdrf:
Sorry to say that we cannot merge the fix until IP-validation is fixed. Going to close this pull within a week because of inactivity

PR #200, opened 2015-11-18
call activators of tools only once

HendrikPeilke:
At the moment activators of tools are called twice upon activation, because the setActive(true)-method of the tool is called from the ToolProxy (setActive(true)) and from the ToolManager (setActiveModalTool(...)).


fgdrf:
@HendrikPeilke please sign (-s) your commit (see [ip-validations|https://locationtech.org/eclipse-webhook/services/status_details.php?id=564c6ce9b7676])
Since it is on a branch, you can "push -f" your work to this branch again.

HendrikPeilke:
That happens if you commit only once in a while. :( Sorry for that. 

Have signed and pushed it, but the checks do not run again, probably because nothing changed but commit sign-off. Do they have to?

fgdrf:
Please reset your branch to the commit before your commit and commit againt
with -s option:
    git reset HEAD^1

HTH


2015-11-18 14:28 GMT+01:00 Hendrik Peilke <notifications@github.com>:

PR #203, opened 2015-12-04
update GeoTools to 14.1, GeoScript to 1.6.0, and dependencies

fgdrf:


jodygarnett:
Thanks Frank, reviewing this now (getting an updated dependency list for jan sprint)

plugins/net.sf.opencsv/META-INF/MANIFEST.MF
```diff
@@ -7,4 +7,4 @@ Bundle-ClassPath: .,lib/opencsv-2.3.jar
 Export-Package: au.com.bytecode.opencsv,
  au.com.bytecode.opencsv.bean
 Eclipse-BundleShape: dir
-Bundle-RequiredExecutionEnvironment: JavaSE-1.6
+Bundle-RequiredExecutionEnvironment: JavaSE-1.7
```
jodygarnett: I agree that java 1.7 is required, I think we should update root pom.xml property <java-version>1.6</java-version>

PR #206, opened 2016-01-11
Fix performance issues on layer refresh

fgdrf:
and avoid re-creation renderer configuration every time.

We have problems if layers having programatically a high refresh intervall. Because of resource changed events the refresh is triggered. Same for Selection Events. We set it programmatically as well to have selections in a table in synch with map selection.

Because of 
```
getRendererCreator().reset(); 
```
if forces to recreate context and configuration. Dependent from the layers are currently on the map it can take a while, especially if there are layers with remote services that responding badly.


fgdrf:
Feedback from Hudson-Instance for this pull request : https://hudson.locationtech.org/udig/job/uDig-PR-NG/27

fgdrf:
Closing this pull because its a mixup of fixes for Classloading performance and Synchronization issues

PR #215, opened 2016-10-25
Fixed resulting pdf page size format

fgdrf:
The resulting PDF page format scaled up and it depends from the selected DPI value. For 72dpi the page size was almost correct, if a user selected 144 DPI the page size doubled up and so on.

The fix is to respect select paper size and scales the image down to the page, resolution is as selected.

### DIFF
--- a/plugins/org.locationtech.udig.printing.ui/src/org/locationtech/udig/printing/ui/pdf/SivecoExportMapToImageWizard.java
+++ b/plugins/org.locationtech.udig.printing.ui/src/org/locationtech/udig/printing/ui/pdf/SivecoExportMapToImageWizard.java
@@ -199,8 +199,7 @@ public class SivecoExportMapToImageWizard extends Wizard implements IExportWizar
                         paper, 
                         10, /*top and bottom margin*/
                         10, /*left and right margin*/
-                        isLandscape,
-                        dpi);
+                        isLandscape);
 	    
         monitor.done();
         
--- a/plugins/org.locationtech.udig.printing.ui/src/org/locationtech/udig/printing/ui/pdf/SivecoExportMapToImageWizard.java
+++ b/plugins/org.locationtech.udig.printing.ui/src/org/locationtech/udig/printing/ui/pdf/SivecoExportMapToImageWizard.java
@@ -30,36 +30,10 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 
-import org.locationtech.udig.catalog.CatalogPlugin;
-import org.locationtech.udig.catalog.ICatalog;
-import org.locationtech.udig.catalog.ID;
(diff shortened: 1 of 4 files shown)


fgdrf:
@egouge What do you think from user perspective? Would you expect to get a different pdf format size at the end if the DPI has been increased? This fix uses the correct format settings for the pdf document (A4 = A4 in pdf format). The resulting pdf e.g. for A4 would fit on a A4 page in printer settings and has not to be rescaled in printer settings.

egouge:
I would want the paper size to stay the same.  I know what what a 8.5x11 piece of paper is but not the implications of DPI (in term of paper size).  So I would suggest keeping the page size the same and scaling the map.

fgdrf:
@egouge could you test the behavior and results of pdf exports. I'm interested in your point of view as you have contact to customers and users of uDig.

In the past the users I know were irritated a bit about the output (doubled DPI results into 4 times bigger image/pdf). The user had to re-scale the image to printer page size that was choosen in the export wizard - and was confused.
HTH

egouge:
@fgdrf This seems to work well in terms of maintaining the page size but I noticed some distortions which I don't think should be there when I print in landscape mode.
landscape
portrait

fgdrf:
@egouge Please correct me if I'm wrong. Are you saying that the first images is made as "screenshot" as it is displayed in map panel, whereas the secound is a screenshot of exported pdf or vice versa? However that would be a blocking issue for this pull request. I'm going to investigate whats wrong ..

egouge:
The second is a screen shot of the map exported to pdf in portrait mode.  The first is the same map exported to pdf in landscape mode.  The only thing I changed was portrait and landscape mode.  Nothing else changed. I will try to attach the actual pdfs when I get the chance.

fgdrf:
@egouge Fixed distortion problems. I stumbled about scale computation problems during export. If the user specify a scale denominator, lets say 700.000 the BoundsStrategy sets it for the RendererMap. After that the bounds are re-calculated. So far so good **but** quering the scale from the ViewportModel again returns a different value than set, e.g. 699999.9999995704 or 700000.0000001818 depenent from selected DPI

However, the computation is another issue we should not adress right here ;)

fgdrf:
And not to forget: The Dialog provides four fields to set margins (top, bottom, left, right) but only two of them are used (leftMarginSpinner and topMarginSpinner).

fgdrf:
Last commit fixes the margin applyment for each direction.

fgdrf:
@moovida Would you please have a look at final changes since your implemented the pdf export initally.

egouge:
@fgdrf this might be my problem but I updated just now and the SivecoExportMapToImageWizard has a compile error.

fgdrf:
Oh my fault, I fixed it and updated last commit.

egouge:
works fine now

PR #218, opened 2017-03-08
Improvements related to bulk feature copy or move between layers

nprigour:
Enhancements include:

1. allow move of features between layers even when there is no strict featureType match (i.e. reordered attribute fields exist). 
2. copy of an attribute field between 2 layers even if the target attribute type does not strictly match the source attribute type based on the existence of proper adapters. 
Included example shp files to test the behavior via the UI.
[test_shp_files.zip](https://github.com/locationtech/udig-platform/files/827288/test_shp_files.zip)

### DIFF
--- a/plugins/org.locationtech.udig.core.tests/src/org/locationtech/udig/core/NumberAdaptableFactory.java
+++ b/plugins/org.locationtech.udig.core.tests/src/org/locationtech/udig/core/NumberAdaptableFactory.java
@@ -1,0 +1,52 @@
+package org.locationtech.udig.core;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.sql.Date;
+import java.sql.Timestamp;
+
+import org.eclipse.core.runtime.IAdapterFactory;
+
+public class NumberAdaptableFactory implements IAdapterFactory {
+	
+	@Override
+	public Object getAdapter(Object adaptableObject, Class adapterType) {
+
+		if (adaptableObject == null) {
+			return null;
+		}
+		
+		if (adaptableObject instanceof Number) {
+			if (adapterType==Long.class) {
+				//System.out.println("Adapting " + adaptableObject.getClass()  + " to Long. Rounding or truncation may occur");
+				return ((Number)adaptableObject).longValue();
+			} else if (adapterType==Integer.class) {
+				//System.out.println("Adapting " + adaptableObject.getClass()  + " to Integer. Rounding or truncation may occur");
+				return ((Number)adaptableObject).intValue();
+			} else if (adapterType==Double.class) {
+				//System.out.println("Adapting " + adaptableObject.getClass()  + " to Double. Rounding may occur");
+				return ((Number)adaptableObject).doubleValue();
+			} else if (adapterType==Short.class) {
+				//System.out.println("Adapting " + adaptableObject.getClass()  + " to Short. Rounding or truncation may occur");
+				return ((Number)adaptableObject).shortValue();
+			} else if (adapterType==Float.class) {
+				//System.out.println("Adapting " + adaptableObject.getClass()  + " to Long. Rounding may occur");
+				return ((Number)adaptableObject).floatValue();
+			} else if (adapterType==BigDecimal.class) {
+				//System.out.println("Adapting " + adaptableObject.getClass()  + " to BigDecimal");
+				return new BigDecimal(adaptableObject.toString());
+			} else if (adapterType==BigInteger.class) {
+				//System.out.println("Adapting " + adaptableObject.getClass()  + " to BigInteger. Rounding may occur");
+				return new BigDecimal(adaptableObject.toString()).toBigInteger();
+			}
+		}
+		
+		return null;
+	}
+
+	@Override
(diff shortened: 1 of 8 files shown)


nprigour:
Hi @fgdrf ,
What is the status of this PR? Is there a reason prohibiting its acceptance?

fgdrf:
@nprigour Actually it's still not merged yet because of lack of time to review it. ;)

Let's finish #279 first, because we will have some follow-up tasks regarding IP-Review (additional picky back CQ's and for GT itself, I guess 14.1 was approved with some exceptions).

Anybody else to review and comment this changeset now?

plugins/org.locationtech.udig.core/src/org/locationtech/udig/core/internal/FeatureUtils.java
```diff
@@ -333,12 +338,36 @@ public int size() {
...
+                    //if not directly assignable then check for an adapter
+                    if (AdapterUtil.instance.canAdaptTo(
+                            source.getAttribute(sourceAttributeName), 
+                            destSchema.getDescriptor(i).getType().getBinding())) {
+                        try {
+                            attributes[i] = AdapterUtil.instance.adaptTo(
+                                    destSchema.getDescriptor(i).getType().getBinding(), 
+                                    source.getAttribute(sourceAttributeName), mon);
+                        } catch (Exception e) {
+                            e.printStackTrace();
```
fgdrf: can we log it to Plugin-Logger rather than just print it?

plugins/org.locationtech.udig.tools.jgrass/src/org/locationtech/udig/tools/jgrass/utils/OperationUtils.java
```diff
@@ -95,11 +95,16 @@ protected void moveFeatures( final Display display, IProgressMonitor monitor, IL
...
         SimpleFeatureType selectedSchema = selectedLayer.getSchema();
         int compare = DataUtilities.compare(toSchema, selectedSchema);
-        if (compare != 0) {
+        boolean needsReOrder = false;
+        if (compare == -1) {
             showMessage(display,
                     Messages.getString("OperationUtils_warning"), Messages.getString("OperationUtils_sametypeproblem"), //$NON-NLS-1$ //$NON-NLS-2$
                     MSGTYPE.WARNING);
             return;
+        } else if (compare == 1){
```
fgdrf: OK, looks good to me. What do you think about a testcase to verify this behavior of [DataUtilities#compare](https://github.com/geotools/geotools/blob/master/modules/library/main/src/main/java/org/geotools/data/DataUtilities.java#L383) to check compatibility after upgrading GeoTools ;)

plugins/org.locationtech.udig.core.tests/src/org/locationtech/udig/core/NumberAdaptableFactory.java
```diff
@@ -0,0 +1,52 @@
+package org.locationtech.udig.core;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.sql.Date;
```
fgdrf: Wondering why java.sql imports are necessary for this class, please re-organize imports. Any chance to provide Unit tests as well?

fgdrf:
@nprigour Can we go through the comments to get it onto master?

plugins/org.locationtech.udig.core/src/org/locationtech/udig/core/internal/FeatureUtils.java
nprigour: done

plugins/org.locationtech.udig.tools.jgrass/src/org/locationtech/udig/tools/jgrass/utils/OperationUtils.java
nprigour: Indeed the result of this method is essential for correct behavior of copy and move. Therefore, I added a new set of tests (see DataUtilitiesTest class)

plugins/org.locationtech.udig.core.tests/src/org/locationtech/udig/core/NumberAdaptableFactory.java
nprigour: A number of unit tests (not exhaustive) added to verify the possibility of adaptation during Feature copy.
Unnecessary imports removed.

nprigour:
Ok @fgdrf  I addressed all of your comments providing some basic unit testing related to the examplary AdaptableFactory that transforms between various types of numbers.

IMHO the feature provided by this PR is one of the most powerful enhancement in udig sdk since it will enable the potential developer to copy and move feature between layers even with completely different types providing that it has defined an appropriate implementation of IAdapterFactory.

nprigour:
Just as a reminder that upgrading to geotools 22.x will affect this enhancement due to https://github.com/geotools/geotools/pull/2525. Probably  this differentiation of behavior will be detected by the added DataUtilities#compare unit test so we will go through it then.

PR #220, opened 2017-03-09
Preference Page added for WMSService

nprigour:
Added a dedicated Preference Page under the Catalog category from WMS Service which enables the configuration of a timeout value while waiting for a response. This option is needed since by default the blocking time was 30 sec which might not be enough when a large area is requested or the connection is slow. 
image


plugins/org.locationtech.udig.catalog.wms/src/org/locationtech/udig/catalog/wms/preferences/WmsPreferenceInitializer.java
```diff
@@ -0,0 +1,24 @@
...
+ * @author Nikolaos Pringouris <nprigour@gmail.com>
+ *
+ */
+public class WmsPreferenceInitializer extends AbstractPreferenceInitializer {
+
+    @Override
+    public void initializeDefaultPreferences() {
+        IPreferenceStore store = WmsPlugin.getDefault().getPreferenceStore();
+       
+        store.setDefault(WmsPreferencePage.WMS_RESPONSE_TIMEOUT, 30);
```
fgdrf: can we define a constant with a semantic to describe what "30" means, e.g. DEFAULT_WMS_RESPONSE_TIMEOUT_IN_SECONDS:

`public static final int DEFAULT_WMS_RESPONSE_TIMEOUT_IN_SECONDS = 30;`

In addition, is it possible to Start uDig with another default (overwrite in configuration) and how?

@nprigour Could you add a screenshout to docs and improve WMS Preferences page (content of http://udig.github.io/docs/user/reference/Preferences.html)  or [Catalog Preferences](https://github.com/locationtech/udig-platform/blob/master/docs/user/en/reference/Catalog%20Preferences.rst)

plugins/org.locationtech.udig.catalog.wms/src/org/locationtech/udig/catalog/wms/preferences/WmsPreferencePage.java
```diff
@@ -0,0 +1,39 @@
...
+	@Override
+	public void init(IWorkbench workbench) {
+		setPreferenceStore(WmsPlugin.getDefault().getPreferenceStore());
+		setDescription("Configuration Options about WMS connections");
+	}
+
+	@Override
+	protected void createFieldEditors() {
+		
+		addField(new IntegerFieldEditor(WMS_RESPONSE_TIMEOUT, "WMS response timeout", getFieldEditorParent()));
```
fgdrf: Could you please externalize the string to properties files as well? So Users could translated it usin e.g. Transifex (https://www.transifex.com/udig/udig/dashboard/)

nprigour:
I have amended the pull request to address your comments regarding translation/internationalization and user documentation.  
Overwrite default value for WMS_RESPONSE_TIMEOUT is not possible (why do we need that really?).  User can change this via the preference page if he experience problems obtaining the WMS response

fgdrf:
@nprigour Overwrite is interesting for automatic deployment without giving the users a checklist with tasks to configure their instance. It was just an idea but I cannot find the link to the discussion we had on the mailing list

Looks good to me, thanks again for your contribution!

PR #208, opened 2016-04-19
New render manager using asynchronous queue for rendering requests to solve problems of legacy design

vitalus:
Code is a direct port from our custom net.refractions.udig - based product.

An idea is to have 1 special thread per map processing  asynchronously coming requests for rendering/repainting a map image.   There might be  many situations for race conditions in legacy rendering logic and queue-based approach solves them.  From any other thread various events safely come to the queue and the special thread "compresses" them into 1 optimized event that whether forces just repainting of cached map image with temporary graphics or requests re-rendering of layers' images or handles rendering of the special selection layer, etc.

Manual test plan:
- add layer to the map
- remove layer from the map
- make selections in vector layers
- change order of layers
- quickly zoom in, pan and add/remove layers at the same time
- change SLD style of vector layer
  ..

all these actions will put events to the queue.

Thread takes raw  events from the queue and compressed them to understand what single action with map image is needed and then accordingly this knowledge delegates to legacy rendering APIs.

Re-evaluation of map image rendering configuration  that was causing many problems in UDIG is now in safe place: new thread that works with asynchronous queue.

PR #224, opened 2017-03-15
fix render metric sorter comparison implementation

nprigour:
fix render metric sorter comparison implementation (`compareMetrics( AbstractRenderMetrics o1, AbstractRenderMetrics o2 )`) to:

- consider both arguments during checking for a preferred renderer
- check both layer and map blackboard settings

 
### DIFF
--- a/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/RenderMetricsSorter.java
+++ b/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/RenderMetricsSorter.java
@@ -97,36 +97,35 @@ public class RenderMetricsSorter implements Comparator<InternalRenderMetrics>, S
         int first = -1;
         int last = 1;
         
-        // check the layer blackboard for a preferred or non-preferred renderer
         // check the layer blackboards for a preferred or non-preferred renderer
-        int perferredMapRenderer = rateUsingBlackboardSettings(o1, o1.getRenderContext().getLayer()
+        int preferredLayerRenderer = rateUsingBlackboardSettings(o1, o1.getRenderContext().getLayer()
                 .getBlackboard());
-        if (perferredMapRenderer == 0) {
-        	perferredMapRenderer = rateUsingBlackboardSettings(o1, o1.getRenderContext().getLayer()
+        if (preferredLayerRenderer == 0) {
+            preferredLayerRenderer = rateUsingBlackboardSettings(o1, o1.getRenderContext().getLayer()
                     .getStyleBlackboard());
         }
-        int perferredMapRenderer2 = rateUsingBlackboardSettings(o2, o2.getRenderContext().getLayer()
+        int preferredLayerRenderer2 = rateUsingBlackboardSettings(o2, o2.getRenderContext().getLayer()
                 .getBlackboard());
-        if (perferredMapRenderer2 == 0) {
-        	perferredMapRenderer2 = rateUsingBlackboardSettings(o2, o2.getRenderContext().getLayer()
+        if (preferredLayerRenderer2 == 0) {
+            preferredLayerRenderer2 = rateUsingBlackboardSettings(o2, o2.getRenderContext().getLayer()
                     .getStyleBlackboard());
         }       
-        if (perferredMapRenderer > perferredMapRenderer2) {
+        if (preferredLayerRenderer > preferredLayerRenderer2) {
             return first;
         }
-        if (perferredMapRenderer < perferredMapRenderer2) {
+        if (preferredLayerRenderer < preferredLayerRenderer2) {
             return last;
         }
         
         // check the map blackboard for a preferred or non-preferred renderer
-        perferredMapRenderer = rateUsingBlackboardSettings(o1, o1.getRenderContext().getMap()
+        int preferredMapRenderer = rateUsingBlackboardSettings(o1, o1.getRenderContext().getMap()
                 .getBlackboard());
-        perferredMapRenderer2 = rateUsingBlackboardSettings(o2, o2.getRenderContext().getMap()
+        int preferredMapRenderer2 = rateUsingBlackboardSettings(o2, o2.getRenderContext().getMap()
                 .getBlackboard());  
-        if (perferredMapRenderer > perferredMapRenderer2) {
+        if (preferredMapRenderer > preferredMapRenderer2) {
             return first;
         }
-        if (perferredMapRenderer < perferredMapRenderer2) {
+        if (preferredMapRenderer < preferredMapRenderer2) {
             return last;
         }
         
--- a/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/RenderMetricsSorter.java
+++ b/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/RenderMetricsSorter.java
@@ -97,12 +97,27 @@ public class RenderMetricsSorter implements Comparator<InternalRenderMetrics>, S
         int first = -1;
         int last = 1;
         
-        // check the layer/map blackboard for a preferred or non-preferred renderer 
-        int perferredMapRenderer = rateUsingBlackboardSettings(o1, o1.getRenderContext().getMap().getBlackboard());
-        if (perferredMapRenderer > 0){
+        // check the layer blackboard for a preferred or non-preferred renderer
+        int perferredMapRenderer = rateUsingBlackboardSettings(o1, o1.getRenderContext().getLayer()
+                .getBlackboard());
+        int perferredMapRenderer2 = rateUsingBlackboardSettings(o2, o2.getRenderContext().getLayer()
+                .getBlackboard());        
+        if (perferredMapRenderer > perferredMapRenderer2) {
             return first;
         }
-        if (perferredMapRenderer < 0){
+        if (perferredMapRenderer < perferredMapRenderer2) {
+            return last;
+        }
+        
+        // check the map blackboard for a preferred or non-preferred renderer
+        perferredMapRenderer = rateUsingBlackboardSettings(o1, o1.getRenderContext().getMap()
+                .getBlackboard());
+        perferredMapRenderer2 = rateUsingBlackboardSettings(o2, o2.getRenderContext().getMap()
+                .getBlackboard());  
+        if (perferredMapRenderer > perferredMapRenderer2) {
+            return first;
+        }
+        if (perferredMapRenderer < perferredMapRenderer2) {
             return last;
         }
         
--- a/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/RenderMetricsSorter.java
+++ b/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/RenderMetricsSorter.java
@@ -99,9 +99,9 @@ public class RenderMetricsSorter implements Comparator<InternalRenderMetrics>, S
         
         // check the layer blackboard for a preferred or non-preferred renderer
         int perferredMapRenderer = rateUsingBlackboardSettings(o1, o1.getRenderContext().getLayer()
-                .getBlackboard());
+                .getStyleBlackboard());
         int perferredMapRenderer2 = rateUsingBlackboardSettings(o2, o2.getRenderContext().getLayer()
-                .getBlackboard());        
+                .getStyleBlackboard());        
         if (perferredMapRenderer > perferredMapRenderer2) {
             return first;
         }
--- a/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/RenderMetricsSorter.java
+++ b/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/RenderMetricsSorter.java
@@ -98,10 +98,19 @@ public class RenderMetricsSorter implements Comparator<InternalRenderMetrics>, S
         int last = 1;
         
         // check the layer blackboard for a preferred or non-preferred renderer
+        // check the layer blackboards for a preferred or non-preferred renderer
         int perferredMapRenderer = rateUsingBlackboardSettings(o1, o1.getRenderContext().getLayer()
-                .getStyleBlackboard());
+                .getBlackboard());
+        if (perferredMapRenderer == 0) {
+        	perferredMapRenderer = rateUsingBlackboardSettings(o1, o1.getRenderContext().getLayer()
+                    .getStyleBlackboard());
+        }
         int perferredMapRenderer2 = rateUsingBlackboardSettings(o2, o2.getRenderContext().getLayer()
-                .getStyleBlackboard());        
+                .getBlackboard());
+        if (perferredMapRenderer2 == 0) {
+        	perferredMapRenderer2 = rateUsingBlackboardSettings(o2, o2.getRenderContext().getLayer()
+                    .getStyleBlackboard());
+        }       
         if (perferredMapRenderer > perferredMapRenderer2) {
             return first;
         }
--- a/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/TiledRendererCreatorImpl.java
+++ b/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/TiledRendererCreatorImpl.java
@@ -322,6 +322,18 @@ public class TiledRendererCreatorImpl implements RendererCreator {
     }
 
     /**
+     * @throws UnsupportedOperationException
+     */
+    public Collection<String> getAvailableRendererIds(Layer layer) {
+    	List<String> result = new ArrayList<String>(); 
+    	Collection<AbstractRenderMetrics> metrics = getAvailableRendererMetrics(layer);
+    	for (AbstractRenderMetrics metric : metrics) {
+    		result.add(metric.getId());
+    	}
+    	return result;
+    }
+    
+    /**
      * 
      * @throws UnsupportedOperationException
      */
--- a/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/TiledRendererCreatorImpl.java
+++ b/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/TiledRendererCreatorImpl.java
@@ -321,8 +321,9 @@ public class TiledRendererCreatorImpl implements RendererCreator {
         throw new UnsupportedOperationException("Renderers don't have information in this system."); //$NON-NLS-1$
     }
 
+
     /**
-     * @throws UnsupportedOperationException
+     * 
      */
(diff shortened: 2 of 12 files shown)


fgdrf:
@nprigour Many Thanks for your patch!

Could you describe the problems in applications with the privious implementation? A short description of the workflow to reproduct would help to understand it and to verify if the fix helps.

nprigour:
this method is used to order renderers applicable for a layer in order to eventually select the best one to be used. There are 2 main issues: 

1. Supposing we have a list of renderers that are allowed for a particular layer. Then the blackboard parameter  RendererCreatorImpl.PREFERRED_RENDERER_ID is used to explicitly define a renderer as preferable over the others (without the need to consider further metrics). The setting of this parameter can be done at the layer blackboard (in order to apply to a particular layer) or at the Map blackboard in order to apply to all layers of a Map (imagine that the setting of this value can be done within a udig layer or Map interceptor). The previous implementation considers only the parameter at the Map level while it should first check the layer blackboard (since it is more common to be set there).    
2. The previous implementation is erroneous in the following sense. The renderer passed as second argument in the method `compareMetrics( AbstractRenderMetrics o1, AbstractRenderMetrics o2 )` is never considered if it is a preferable one. Even if its id is set in  RendererCreatorImpl.PREFERRED_RENDERER_ID it will not be considered during comparison and we will eventually resort to a comparison based on defined metrics using the `private double rate( InternalRenderMetrics metrics )` method. It will only be considered by change in the case that the renderer provided as 1st argument is set as RendererCreatorImpl.LAST_RESORT_RENDERER_ID.

fgdrf:
@vitalus Would you please have a look at this changeset, I assume your are very familiary with the Renderer-Subsystem (see open pull #208). Many thanks

nprigour:
@fgdrf  in order to proceed with this one I have added some extra classes that can be used for testing. You should just include org.locationtech.udig.project.tests plugin (and its dependencies) when launching udig.
Then you can use the extra menu item option provided on top level menu (see screenshot) to verify and compare with the previous implementation:

- correct usage of RendererCreatorImpl.PREFERRED_RENDERER_ID at map and layer blackboard where layer setting takes always precedence,
- correct usage  of RendererCreatorImpl.LAST_RESORT_RENDERER_ID at map and layer blackboard where layer setting takes always precedence
- behavior of metrics sorting in the absence of the above

image

nprigour:
Status of these PR? It is almost a year old!

plugins/org.locationtech.udig.project.tests/META-INF/MANIFEST.MF
```diff
@@ -24,3 +24,5 @@ Require-Bundle: org.junit;bundle-version="[4.0.0,5.0.0)",
  org.easymock
 Bundle-ActivationPolicy: lazy
 Bundle-RequiredExecutionEnvironment: JavaSE-1.7
+Import-Package: org.locationtech.udig.project.ui,
```
fgdrf: IMO we should keep UI contributions seperated from core tests. What do you think about creating a new bundle org.locationtech.udig.project.ui.tests which has a dependency to org.locationtech.udig.project.ui and provides Eclipse menue and command contributions.

It should not be part of a standard eclipse feature to avoid getting it into product. On the other hand it might be worth to have it in sdk

plugins/org.locationtech.udig.project.tests/src/org/locationtech/udig/project/command/RendererSetterHandler.java
```diff
@@ -0,0 +1,431 @@
...
+	        rating += userAppearanceMetric * WEIGHT_USER_APPEARANCE_METRIC;
+	        rating += latencyMetric * WEIGHT_LATENCY_METRIC;
+	        rating += drawingTimeMetric * WEIGHT_DRAWING_TIME_METRIC;
+	        rating += resolutionMetric * WEIGHT_RESOLUTION_METRIC;
+	        rating += multiLayerMetric * WEIGHT_MULTILAYER_METRIC;
+	        rating += scaleRangeMetric * WEIGHT_SCALERANGE_METRIC;
+	        
+	        return rating;
+	        
+//	        int rating = rateAppearance(metrics);
```
fgdrf: I assume it was for testing, we should clean up commented code here

plugins/org.locationtech.udig.project.tests/src/org/locationtech/udig/project/command/RendererSetterHandler.java
```diff
@@ -0,0 +1,431 @@
...
+import org.locationtech.udig.project.ui.internal.ApplicationGISInternal;
+
+
+/**
+ * Test handler that provides a way to check behavior of {@link RenderMetricsSorter}.
+ * 
+ * @author Nikolaos Pringouris <nprigour@gmail.com>
+ *
+ */
+public class RendererSetterHandler extends AbstractHandler {
```
fgdrf: This class should be refactored to suggested project.ui.tests bundle

plugins/org.locationtech.udig.project.tests/src/org/locationtech/udig/project/internal/render/impl/renderercreator/SingleFeatureRenderMetricsFactory2.java
```diff
@@ -0,0 +1,75 @@
+/* uDig - User Friendly Desktop Internet GIS client
+ * http://udig.refractions.net
+ * (C) 2004, Refractions Research Inc.
```
fgdrf: Please use year of contribution

plugins/org.locationtech.udig.project.tests/src/org/locationtech/udig/project/internal/render/impl/renderercreator/SingleFeatureRenderMetricsFactory2.java
```diff
@@ -0,0 +1,75 @@
...
+import org.locationtech.udig.project.render.IRenderContext;
+import org.locationtech.udig.project.render.IRenderMetricsFactory;
+import org.locationtech.udig.project.render.IRenderer;
+
+/**
+ * For testing.  Creates a normal Renderer.  Accepts resources that resolve to FeatureSource objects.
+ * 
+ * @author Nikolaos Pringouris <nprigour@gmail.com>
+ */
+public class SingleFeatureRenderMetricsFactory2 implements IRenderMetricsFactory {
```
fgdrf: Whats the different between SingleFeatureRenderMetricsFactory and SingleFeatureRenderMetricsFactory2?

plugins/org.locationtech.udig.project.tests/src/org/locationtech/udig/project/internal/render/impl/renderercreator/SingleFeatureRenderer2.java
```diff
@@ -0,0 +1,33 @@
...
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.locationtech.udig.project.internal.render.impl.RendererImpl;
+import org.locationtech.udig.project.render.RenderException;
+
+/**
+ * Simulates a simple renderer that renders a single class.
+ * 
+ * @author Nikolaos Pringouris <nprigour@gmail.com>
+ */
+public class SingleFeatureRenderer2 extends RendererImpl {
```
fgdrf: Same here, ist it to have two different type of RendererImpl? What about using the same class twice for two different renderer-Extensions in plugin.xml?

plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/TiledRendererCreatorImpl.java
```diff
@@ -321,6 +321,18 @@ public SelectionLayer findSelectionLayer( ILayer targetLayer ) {
         throw new UnsupportedOperationException("Renderers don't have information in this system."); //$NON-NLS-1$
     }
 
+    /**
+     * @throws UnsupportedOperationException
```
fgdrf: Where is a UnsupportedOperationException thrown?

plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/RenderMetricsSorter.java
```diff
@@ -97,12 +97,36 @@ private int compareMetrics ( InternalRenderMetrics o1, InternalRenderMetrics o2
...
+        // check the map blackboard for a preferred or non-preferred renderer
+        perferredMapRenderer = rateUsingBlackboardSettings(o1, o1.getRenderContext().getMap()
+                .getBlackboard());
+        perferredMapRenderer2 = rateUsingBlackboardSettings(o2, o2.getRenderContext().getMap()
+                .getBlackboard());  
+        if (perferredMapRenderer > perferredMapRenderer2) {
             return first;
         }
-        if (perferredMapRenderer < 0){
+        if (perferredMapRenderer < perferredMapRenderer2) {
```
fgdrf: This is the same check like in line 117. I guess it's a different comparision (StyleBlackboard vs. Blackboard) and code would be better readable if the name of the variables have a semantic.
However, looks like duplicated code/checks I don't understand 😬

plugins/org.locationtech.udig.project.tests/src/org/locationtech/udig/project/internal/render/impl/renderercreator/SingleFeatureRenderMetricsFactory2.java
nprigour: OK I will adapt header comment in all new classes

plugins/org.locationtech.udig.project.tests/src/org/locationtech/udig/project/internal/render/impl/renderercreator/SingleFeatureRenderMetricsFactory2.java
nprigour: actually the contain an AbstractRenderMetrics with different values for certain metrics (latency & timeToDrawMetric) so that we are able to check explicit assignment via StyleBoardParameter and automatic via render metric

plugins/org.locationtech.udig.project.tests/src/org/locationtech/udig/project/internal/render/impl/renderercreator/SingleFeatureRenderer2.java
nprigour: Actually it will make easier to distinguish the preferred renderer during testing via the extra menu option provided since class names are used for display and select 
image

plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/TiledRendererCreatorImpl.java
nprigour: You are right I will remove it

plugins/org.locationtech.udig.project.tests/META-INF/MANIFEST.MF
nprigour: actually there is plugin org.locationtech.udig.project.tests.ui. 
Maybe I can use this for defining the menu and command contribution since it already contains other eclipse UI elements.
What do you think @fgdrf ?

nprigour:
Hi @fgdrf ,
if you do not have an objection, before I apply (wherever possible) you suggested comments and since this PR is pretty old and has conflicts with the master branch, I will first try to rebase  master branch to it and forcibly push it so we have a more easy job while merging and closing it.
Just confirm whether you agree or not!

plugins/org.locationtech.udig.project.tests/src/org/locationtech/udig/project/internal/render/impl/renderercreator/SingleFeatureRenderMetricsFactory2.java
fgdrf: Does it makes sense to rename these classes for better readability? Like Latency.. TimeToDrawRendereMetricsFactory

fgdrf:
Totally agree, if so we would get Feedback from Jenkins for this pull request 😄  which would help to verify if everything else is still valid

nprigour:
I have addressed almost all of your comments  plus improve a little bit the test handler in order to display only applicable  renderer metrics in its drop down combo. The changes have been forcibly pushed overwriting the previous changes

fgdrf:
Hmm, I thought you had rebased to latest from master already and I was wondering why 14.1 GeoTools dependences were downloaded...
I'll test with now this environment but rebase would make it easier to switch between branchesn ;)

Many Thanks

fgdrf:
Looks like if I run from within Eclipse with project.test.ui bundle the map cannot render anything. I assume its because of the following exception:

```
!ENTRY org.locationtech.udig.project.tests 2 0 2018-05-17 10:55:29.073
!MESSAGE Error processing extension java.lang.RuntimeException
!STACK 0
java.lang.RuntimeException
	at org.locationtech.udig.project.internal.render.impl.renderercreator.BadRenderMetricsFactory.createMetrics(BadRenderMetricsFactory.java:105)
	at org.locationtech.udig.project.internal.render.impl.InternalRenderMetricsFactory.createMetrics(InternalRenderMetricsFactory.java:47)
	at org.locationtech.udig.project.internal.render.impl.RendererExtensionProcessor.process(RendererExtensionProcessor.java:91)
	at org.locationtech.udig.core.internal.ExtensionPointUtil.process(ExtensionPointUtil.java:77)
	at org.locationtech.udig.project.internal.render.impl.RendererCreatorImpl.initFactories(RendererCreatorImpl.java:617)
	at org.locationtech.udig.project.internal.render.impl.RendererCreatorImpl.initRenderMetrics(RendererCreatorImpl.java:328)
	at org.locationtech.udig.project.internal.render.impl.RendererCreatorImpl.createConfiguration(RendererCreatorImpl.java:227)
	at org.locationtech.udig.project.internal.render.impl.RendererCreatorImpl.getConfiguration(RendererCreatorImpl.java:630)
	at org.locationtech.udig.project.ui.internal.RenderManagerDynamic.validateRendererConfiguration(RenderManagerDynamic.java:433)
	at org.locationtech.udig.project.ui.internal.RenderManagerDynamic.refresh(RenderManagerDynamic.java:270)
	at org.locationtech.udig.project.ui.internal.RenderManagerAdapters$1.refreshDirtyArea(RenderManagerAdapters.java:85)
	at org.locationtech.udig.project.ui.internal.RenderManagerAdapters$1.notifyChanged(RenderManagerAdapters.java:65)
	at org.eclipse.emf.common.notify.impl.BasicNotifierImpl.eNotify(BasicNotifierImpl.java:374)
	at org.locationtech.udig.project.internal.render.impl.ViewportModelImpl.fireNotification(ViewportModelImpl.java:905)
	at org.locationtech.udig.project.internal.render.impl.ViewportModelImpl.setBoundsInternal(ViewportModelImpl.java:435)
	at org.locationtech.udig.project.internal.render.impl.ViewportModelImpl.setBounds(ViewportModelImpl.java:380)
	at org.locationtech.udig.project.internal.render.impl.ViewportModelImpl.zoomToBox(ViewportModelImpl.java:959)
	at org.locationtech.udig.project.internal.render.impl.ViewportModelImpl$1.run(ViewportModelImpl.java:888)
	at java.lang.Thread.run(Thread.java:745)
```

I can see which RendererMetrics are applicatable and which not. I'm wondering why 
```-  org.locationtech.udig.render.internal.feature.shapefile.ShapefileFeatureMetrics@2a7b5925 --> (rate:0.39236441647597253)
``` isn't applicateable while I've added a ShapefileLayer? Can you explain this?

**Comment by nprigour** [comment]
Actually @fgdrf  the intention of the test is only to check the proper sorting of renderers and how the setting of keys (i.e. PREFERRED_RENDERER_ID) affect the selected renderer. Both the **LowLatency** and **HighLatency** renderers actually do nothing (their render() method is empty). And since LowLatency by default takes precedence over all others (it has the lower computed rate) that is why you do not see anything render.
If you load a shapefile layer and check with the test button "Renderer Testing" by selecting the layer then you will probably see  in the texArea below something like the following :
```
APPLICABLE RendererMetrics FOR LAYER :Layer ( name: pst):
 1.  org.locationtech.udig.project.internal.render.impl.renderercreator.LowLatencyFeatureRenderMetricsFactory$LowLatencyFeatureRendererMetrics@5cfa9e8a --> (rate:0.45000000000000007)
2.  org.locationtech.udig.project.internal.render.impl.renderercreator.HighLatencyFeatureRenderMetricsFactory$HighLatencyFeatureRendererMetrics@218fc40d --> (rate:0.45033333333333336)
3.  org.locationtech.udig.render.internal.feature.basic.BasicFeatureMetrics@22942eff --> (rate:0.45512688734030204)
```
which means that only these 3 renderers are available for rendering the layer (if you check with a mapGraphic or Grid layer then you will see something else)

Now by the buttond and the combo provide above you can affect the renderer that will eventually be used. For example if you set as preferred the **org.locationtech.udig.render.internal.feature.basic.BasicFeatureMetrics** which is the default  and then close and reopen the map you will see the shapefile properly rendered.

Now regarding you question about why `org.locationtech.udig.render.internal.feature.shapefile.ShapefileFeatureMetrics` is not applicable for the shapefile I suppose it has to do with whether its `canRender(IRenderContext context)` method resolves to true or false. From a quick test I did with a shapefile layer it seems that it resolves to false (in the code below the execution never enters the if statement!!) :
```

 public boolean canRender(IRenderContext context) {
        try {
            IGeoResource geoResource = context.getGeoResource();
            if( geoResource.canResolve(ShapefileDataStore.class)){
                FeatureSource featureSource=geoResource.resolve(FeatureSource.class, null);
                
                boolean notAView = !(featureSource instanceof DefaultView);
                boolean isAShapefile = (featureSource.getDataStore() instanceof ShapefileDataStore);
                return notAView && featureSource!=null && isAShapefile; 
            }
        }
        catch( Throwable t ) {
        }
        return false;
    }
```

fgdrf:
> Now regarding you question about why org.locationtech.udig.render.internal.feature.shapefile.ShapefileFeatureMetrics is not applicable for the shapefile I suppose it has to do with whether its canRender(IRenderContext context) method resolves to true or false. From a quick test I did with a shapefile layer it seems that it resolves to false (in the code below the execution never enters the if statement!!)

Looks like another problem under the hood that releated to your changes but should addressed later.

fgdrf:
@nprigour Sorry for confusion, I've merged before getting feedback from https://ci.eclipse.org/udig/job/uDig-PR/

Can you please create another pull request, there we have to address regressions from https://ci.eclipse.org/udig/job/uDig-master-CI/124/

nprigour:
Hi @fgdrf, 
- In my opinion we should mark this as an improvement.
- Now I am not sure I fully understand what you want . Do you want to create another PR as a replicate of #224?
- the test failures you mention in the jenkins link are due to the #224 introduced changes?

fgdrf:
cherry-picked your changes and created another pull request, for Details see #288

PR #226, opened 2017-03-20
[pdf] remember export settings for pdf wizard page

fgdrf:
This pull request improves the PdfExportWizard page. Whenever the user exports a pdf the settings are stored in a PreferenceStore. Stored settings such as 
* dpi,
* landscape mode,
* page format, and
* margin settings for
   * top
   * bottom
   * left and
   * right

are requested if the user exports the map to pdf again. So the user doesn't have to re-configure the same settings again and again.

In addition refactored properties naming (e.g. lower to bottom) 
 

fgdrf:
Tested in a running application to export to pdf again, settings were still present. Restarted application without deleting the workspace and settings haven't changed since restart

PR #227, opened 2017-03-21
cleanup non-functional & un-maintained csw modules

fgdrf:
The catalog module was listed in plugins/pom.xml (and packaged on each build) the bundle isn't referenced in any feature and therefor not present in uDig final product.

In addition relevant code fragements were commented out whicht makes non-functional.

the libs/scala-csw-client module is a scala project wasn't build for years with scala either.

However, if somebody is willing to maintain it, let's know. Is there anybody out there who used it in production? 

Please feel free to comment and vote to:
:+1: delete
:-1: keep

Thanks in advance


fgdrf:
@jesseeichar @jodygarnett Opinions?

jodygarnett:
👎 we should lighten up

fgdrf:
The question is: Who can maintain this and when? 

I would suggest to create another GIT-Repository for the scala-csw-code and reference the libs like other 3rd-party libs such as GeoTools

On the other hand : Does GeoTools provide Implementation for CSW and how can we use this in uDig? How has the time to investigate?

I'd like to avoid keeping unused code in uDig Repository and build over and over again. In addition I couldn't find any help in the docs for Catalog Service Integration.

So : Is there a wider interest to support this feature in the future? or should we make a cut unless somebody is willing to maintain it?

jodygarnett:
GeoTools does not have a CSW client; it has some bindings for publishing but as far as I know the parsing bindings needed for a CSW client have not been written.

It is one of those two 2-3 week projects that needs funding to be successful.

PR #228, opened 2017-05-12
fix for locationtech bug 140

nprigour:
A fix for bug 140 (see https://bugs.locationtech.org/show_bug.cgi?id=140). It seems that for ShapefileDataStore the transaction mode is not set to AutoCommit by default. 

### DIFF
--- a/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/impl/UDIGFeatureStore.java
+++ b/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/impl/UDIGFeatureStore.java
@@ -154,7 +154,8 @@ public class UDIGFeatureStore implements FeatureStore<FeatureType,Feature>, UDIG
         EditManager editManager = (EditManager) layer.getMap().getEditManager();
         Transaction transaction = editManager.getTransaction();
         
-        if (wrapped.getTransaction() == Transaction.AUTO_COMMIT) {
+        if (wrapped.getTransaction() == null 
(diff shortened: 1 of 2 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=140

nprigour:
Can you explain why the pull request failed. The bug refers to locationtech bug (for udig) and has nothing to do with the eclipse issue you mention above.
The past months I had created a number of pull request for udog without any such issue.
If an ECA is really needed (until now I was informed that I need to be a commiter only in the case the changes affected more than 1000 lines of code) can you advice on how to do it.

fgdrf:
I'm wondering about the result of the ip-validation. It seems it changed, I'm going to clarify with Eclipse webmaster. In addition we raised a bug for wrong link generation for locationtech github projects  https://bugs.locationtech.org/show_bug.cgi?id=75

fgdrf:
@nprigour Do you have an Eclipse account? if so, please try to login and "sign" the ECA, this may help. If you don't have an account, I'll try to find out what we need to get it approved.

nprigour:
@fgdrf I just signed the ECA. Please advise on next actions

fgdrf:
@jodygarnett Could you please have a look at? From my point of view and what the Bug states it looks like patch fixes it. Whenever a layer haven't had a transaction, a new one is introduced. This especially happens if a target layer is hit by drag&drop operations or even edits on the Table while no edit tools were involved.

👍  to merge it

PR #229, opened 2017-05-17
minor fix for duplicate legend entries for color ramp rasters

egouge:
Fix for legend issue shown in attached image.
legendfix

### DIFF
--- a/plugins/org.locationtech.udig.legend/src/org/locationtech/udig/legend/ui/ColorMapLegendCreator.java
+++ b/plugins/org.locationtech.udig.legend/src/org/locationtech/udig/legend/ui/ColorMapLegendCreator.java
@@ -205,7 +205,7 @@ public class ColorMapLegendCreator {
 
 			String l1 = entry.getLabel();
 			String l2 = prevEntry.getLabel();
-			if (i <= entries.length - 3){
+			if (i < entries.length - 1){
 				if (l2 == null){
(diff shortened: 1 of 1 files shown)


plugins/org.locationtech.udig.legend/src/org/locationtech/udig/legend/ui/ColorMapLegendCreator.java
```diff
@@ -205,7 +205,7 @@ public ImageData getImageData() {
 
 			String l1 = entry.getLabel();
 			String l2 = prevEntry.getLabel();
-			if (i <= entries.length - 3){
+			if (i < entries.length - 1){
```
egouge: Q: What's the reason/use-case that String[] text can have length of 1 or the length of two?
A: From what I remember - it was the only way I could "easily" get "gradient" styles to appear in the correctly without substantially rewriting other parts of the Legend Graphic.   
In the case shown above there are four labels (0, 3, 6, 8) but in terms of the colors (and legend entries) there are only three: white to light green [0-3]; light green to medium green [3-6]; and medium green to dark green [6,8].  So in order to get all the labels to show up one of those three legend entries needed to have two labels.
I am sure there were other ways of doing this, but this is what I came up with at the time.

fgdrf:
@egouge Thank you very much for the fix and the continuous support! Much appreciated

PR #230, opened 2017-05-25
provide initialization of oldValue array property in SetAttributesCommand

nprigour:
oldValue array is not initialized during command construction time resulting in a NPE when trying to set values to an array element during run() or rollback method 

### DIFF
--- a/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/commands/edit/SetAttributesCommand.java
+++ b/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/commands/edit/SetAttributesCommand.java
@@ -21,7 +21,7 @@ import org.locationtech.udig.project.command.UndoableMapCommand;
 import org.locationtech.udig.project.command.provider.EditFeatureProvider;
 import org.locationtech.udig.project.command.provider.EditLayerProvider;
 import org.locationtech.udig.project.internal.Messages;
-
+import org.apache.commons.lang.Validate;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.geotools.data.FeatureStore;
 import org.geotools.factory.CommonFactoryFinder;
@@ -34,7 +34,7 @@ import org.opengis.filter.Id;
 
 /**
  * This command modifies an attribute of the current editFeature(the victim that is currently
- * edittable).
+ * editable).
  * 
  * @author jeichar
  * @since 0.3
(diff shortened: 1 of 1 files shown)


plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/commands/edit/SetAttributesCommand.java
```diff
@@ -60,6 +60,7 @@ public SetAttributesCommand( IBlockingProvider<SimpleFeature> feature, IBlocking
             Object value[] ) {
         this.xpath = xpath;
         this.value = value;
+        this.oldValue = new Object[value.length];
```
fgdrf: Maybe its worth to check for null of value here. Looks like the only caller is EditFeature where value is set. However this is an implementation detail but from javadoc constructor argument description it could be possible if somebody like to implement to "clean"/"empty" attributes.

I would suggest to check for null in both constructors or even delegate from the second to the first one using:
```
public SetAttributesCommand( String xpath[], Object value[] ) {
        this(new EditFeatureProvider(this), new EditLayerProvider(this), xpath, value); 
}
```

nprigour: Hi @fgdrf, 
Delegation the way you specify it is not possible since  we cannot refer to 'this' while explicitly invoking a constructor. 
Regarding null value check what can be done is to put an Assert.isNotNull for both xpath[] and value[] at the beginning of each constructor, since providing null values for these parameters is more or less meaningless for this specific command and should be not permitted (the subsequent execution of the run method will fail). 
If a user wants to clean/empty attributes he should set a value[] array with the proper length and null values for each array element desired to be cleaned 
If this will this suffice your request I will add the Assertion checks.

fgdrf: > Delegation the way you specify it is not possible since we cannot refer to 'this' while explicitly invoking a constructor.

Oh yes, haven't seen this. However, the first Constructor is never used in codebase. I'll provide a pull request to eliminate it.

I'm wondering a bit why value.length is used while the loop 
```
		// for loop to get the old values
		for(int i = 0; i < xpath.length; i++){
		    oldValue[i] = feature2.getAttribute(xpath[i]);
		}
```

uses `xpath.lenght`. I assume that object-Array might have same length but who knows ;)

fgdrf: see pull https://github.com/nprigour/udig-platform/pull/10 for validation. I would merge afterwards..

fgdrf: Going to change the pull ;)

nprigour:
I just rebased and merged your pull request

PR #231, opened 2017-05-25
improve check for jdbc catalog IDs

nprigour:
Minor improvement so that ID.isJDBC() method also checks the URL protocol field to determine if ID refers to a database 

### DIFF
--- a/plugins/org.locationtech.udig.catalog/src/org/locationtech/udig/catalog/ID.java
+++ b/plugins/org.locationtech.udig.catalog/src/org/locationtech/udig/catalog/ID.java
@@ -468,7 +468,7 @@ public class ID implements Serializable {
      * @return true if ID refers to a database (ie is a jdbc url)
      */
     public boolean isJDBC() {
-        return id.startsWith("jdbc"); //$NON-NLS-1$
+        return id.startsWith("jdbc") || (url != null && url.getProtocol().contains("jdbc")) ; //$NON-NLS-1$ //$NON-NLS-2$
         // if( url != null){
         // String PROTOCOL = url.getProtocol();
         // String HOST = url.getHost();


plugins/org.locationtech.udig.catalog/src/org/locationtech/udig/catalog/ID.java
```diff
@@ -468,7 +468,7 @@ public boolean isWFS() {
      * @return true if ID refers to a database (ie is a jdbc url)
      */
     public boolean isJDBC() {
-        return id.startsWith("jdbc"); //$NON-NLS-1$
+        return id.startsWith("jdbc") || (url != null && url.getProtocol().contains("jdbc")) ; //$NON-NLS-1$ //$NON-NLS-2$
```
fgdrf: @nprigour Can you describe a scenario if the id wouldnt start with jdbc while protocol could be jdbc? I'm not sure but if the first is true the second statement isn't required. If the first is false the second should either, shouldn't it?

nprigour: ID provides various constructors which do not necessary guarantee that id value is always derived from the URL. Moreover the GeoResource implementation provided for  different databases do not directly use the jdbc URL passed to a java JDBC connection. To give an example with such IDs produced fot Postgres and Mysql:
- Postgres --> example id value **jdbc.postgis**://postgres@pc06543:5432/db_name#testbed
- Mysql --> example id value **mysql.jdbc**://user:psswd@localhost:3306/db_name#test_1 (see also attached screenshot)
In the mysql case it is evident that id does not start with _jdbc_
image

fgdrf:
@naprigour Thanks for explanation and the fix!

PR #232, opened 2017-05-25
improve discovery of all geometry types during when adding a Mysql Host layer

nprigour:
when adding a new Layer through the top level menu and selecting MySql Host as data source (Layer-->add -->MySql Host) after selection of the appropriate DB schema, only tables with a spatial column of generic Geometry are visible. 
The provided pull request improves this by making visible also tables containing a spatial column of all possible geometry types ("point", "multipoint", "line", "linestring", "polygon", "multipolygon") so that they can be added to the CatalogView 


nprigour:
Add attachement to explain improvement
[pull_request232.pdf](https://github.com/locationtech/udig-platform/files/1029254/pull_request232.pdf)

fgdrf:
Sounds very reasonable and patch looks quite good. 👍 

It would be great to have the third screenshot (see attachment pull_request232.pdf) or something similiar on a **new** page in the docs down below http://udig.github.io/docs/user/tasks/Using%20the%20Catalog.html e.g. "Adding Features from MySQL"

@nprigour Can you help here either?

nprigour:
sphinx documentation updated

PR #234, opened 2017-05-31
fix for locationtech bug 167

nprigour:
fix for locationtech bug 167 (https://bugs.locationtech.org/show_bug.cgi?id=167)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=167

nprigour:
@fgdrf there was a mistake in my code. Line 123 should check for not NULL
if (ds.getDataSource() == null) {  --> if (ds.getDataSource() != null) { 
We should revert and fix the code to the pull request before committing

fgdrf:
Fixed it with commit https://github.com/locationtech/udig-platform/commit/482c4a7dca2f912e1757dba79e2048ea40c3bea9

PR #235, opened 2017-06-01
fix for location udig bug-166

nprigour:
fix for location udig bug-166
(https://bugs.locationtech.org/show_bug.cgi?id=166) related to Map
restoration from file

### DIFF
--- a/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/impl/ProjectImpl.java
+++ b/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/impl/ProjectImpl.java
@@ -10,6 +10,8 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 
+import org.apache.commons.lang.StringUtils;
+
 import org.locationtech.udig.project.command.Command;
 import org.locationtech.udig.project.command.CommandManager;
 import org.locationtech.udig.project.internal.Messages;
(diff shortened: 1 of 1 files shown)


nprigour:
It seems that the provided commit fixes the problem in windows OS.
@fgdrf are you able to check it in other operating systems (i.e. linux)

fgdrf:
See my comments on the BUG 166

nprigour:
Hi  @fgdrf,
In order to assist me in debug:

1. temporary apply the attached projectImpl_debug.patch (adds some system.out prints for debug purpose)
1. delete all  .umap files from disk
1. start udig (create a couple of Maps) and exit 
1. restart udig
1. collect the print out results and send them to me by email. 

[projectImpl_debug.zip](https://github.com/locationtech/udig-platform/files/1046811/projectImpl_debug.zip)

fgdrf:
@nprigour I tested today again and I was wondering that my build was quite fast :) Now I found out that I started an older compiled product than expected because I missed to use `-Pproduct` for the maven build.

Finally I can confirm, that your fix works on MacOSX. Therefore I'm going to squash commits soon

fgdrf:
Just to clarify : Whenever a project resource (e.g. a Map) is deleted using the context menu in the layers view the resource on the filesystem is **not** deleted. Files of deleted resources are still present and are not removed closing uDig application. This is possible another issue ..

nprigour:
@fgdrf shouldn't we also check it in a linux environment before accepting it?

Concerning the deletion of resources in the file system it is another thing which however is to a large extend related to the solution we provide to this one. I have filed a seperate bug report on that (see https://bugs.locationtech.org/show_bug.cgi?id=168) and will soon provide a fix via a pull request provided that the present pull request is tested in all major platforms (windows, macOS, linux).

nprigour:
verified successfully in Windows & Linux (Ubuntu) environment

PR #237, opened 2017-06-06
Map project file not deleted from file system fix locationtech [bug-168]

nprigour:
Provides a fix for locationtech [bug-168](https://bugs.locationtech.org/show_bug.cgi?id=168) related to Map file deletion from disk. 
Pull request #235 must be applied before accepting this

### DIFF
--- a/plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/actions/Delete.java
+++ b/plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/actions/Delete.java
@@ -47,6 +47,7 @@ import org.locationtech.udig.ui.PlatformGIS;
 
 import org.eclipse.core.runtime.IAdaptable;
 import org.eclipse.core.runtime.Platform;
+import org.eclipse.emf.common.util.URI;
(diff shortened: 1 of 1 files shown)


fgdrf:
Does the change fix [UDIG-1974](https://osgeo-org.atlassian.net/browse/UDIG-1974)?

fgdrf:
Verified ✔️  that it works fine on windows, project-files (*.umap) are deleted. I'm wondering about log-entries for a deleted map, if it was created within the same session:
`!ENTRY org.locationtech.udig.project 1 0 2017-08-23 10:56:37.662`
`!MESSAGE Not saving file://<path-to-workspace>\project.udig\Map.umap empty contents`

Is it worth to have feedback like this or does it makes sense to avoid this logging-entry to avoid confusions?

nprigour:
I would say that [UDIG-1974](https://osgeo-org.atlassian.net/browse/UDIG-1974) is more related to [locationtech-bug-166](https://bugs.locationtech.org/show_bug.cgi?id=166) which is addressed by pull request #235. 

Concerning the logging I do not think it is completely useless. It is logged at INFO level and may also occur in the case a new project exists without a map or other resource beneath. I opt for keeping it as is.

Issue #247, opened 2017-07-18
introduce enablement of move up/down in layers view


fgdrf:
Dependent from the current selection in LayersView it sometimes doesn't make sense to allow move up/move down of the currently selected layers. For example, if the layer is on top of all others the move up action should be disabled.

![movedown](https://user-images.githubusercontent.com/644229/28308635-1e86d0ac-6ba7-11e7-9275-914c20dfefea.PNG)

![moveup](https://user-images.githubusercontent.com/644229/28308636-1e89177c-6ba7-11e7-9689-996f66b13c6f.PNG)

This fix is related to #240

fgdrf:
@nprigour Thanks again for providing a fix to fix this issue

PR #248, opened 2017-07-18
fixes enablement of layer move up/down actions

fgdrf:
see issue #247

Now the selection is used to enable/disable Move Up/Down Action.

grafik

grafik

If there are Layers in between, that are not selected, it's possible to move up/down

grafik

As a result of **Move Up** the selection it would look like this:

grafik


fgdrf:
@egouge Is this interesting for SMART, isn't it?

nprigour:
Nice improvement  👍  Why is it not yet integrated in the master branch?

fgdrf:
@nprigour, Thanks!

I guess there is a little issue left. Assume the user has three selected layers such as 
grafik

and clicks on the move down action. The first move is fine:
grafik

while the second mixes up the two lower once:
grafik

Not sure if it is really a bug but kind of confusing, isn't it?

nprigour:
added a pull request to your local branch in relation to this. If you agree you can merge it and proceed with integration in master

fgdrf:
I'm going to test your contribution. Thanks for collaboration! Appreciate your help here!

PR #252, opened 2017-09-12
consider system level proxy settings

nprigour:
Improvement consists of:
1. checking the possibility to retrieve proxy setting from system level properties 
2. provide feedback if configuration ini file is not found


plugins/org.locationtech.udig.ui/src/org/locationtech/udig/internal/ui/UiPlugin.java
```diff
@@ -494,27 +504,30 @@ public static Properties getProxySettings() throws FileNotFoundException, IOExce
...
-            if (line.matches(".*D" + RuntimeFieldEditor.PROXYPORT + ".*")) {
-                String proxyPort = line.split("=")[1].trim();
-                properties.put(RuntimeFieldEditor.PROXYPORT, proxyPort);
-            }
-            if (line.matches(".*D" + RuntimeFieldEditor.PROXYNONHOSTS + ".*")) {
-                String proxyNonHosts = line.split("=")[1].trim();
-                // remove quotes if there are
-                proxyNonHosts = proxyNonHosts.replaceAll(PROXYQUOTES, "");
-                properties.put(RuntimeFieldEditor.PROXYNONHOSTS, proxyNonHosts);
+            properties.put(RuntimeFieldEditor.PROXYHOST, System.getProperty(RuntimeFieldEditor.PROXYHOST));
```
fgdrf: I hope I understood it correctly, does it means 
* that System properties are returned for Proxy-configuration if there is no ini-file present
* that configured proxy-condigurations are read from ini file if present?

What if proxy configuration isn't set in ini-file. Does't it mean for callers of  getProxySettings() that there is no proxy-configuration? What do you think about returning system properties if ini file doen't contain any proxy-configuration?

What happens if system Properties are not set either?

nprigour: The bahavior including the latest commit is as follows:

1. The eclipse preference storage is 1st contacted and if proxy information is found then it is also set to the .ini file (if present).
2. Supposing that no info exist in preference storage (common case at first startup) then the .ini file  is looked up for proxy related information. If not present then the System. properties is looked up otherwise no proxy info is set.
3. Upon "reset to default" again the the .ini file  is looked up and if nothing is found there then the System.properties is searched. If system Properties are not set either then obviously the proxy is not initialized (null  or empty values)

PR #253, opened 2017-09-12
fix NPE during initial loading of ProjectExplorer

nprigour:
During initial loading of ProjectExplorer view a NPE may occur if retrieved IProject element resolves to null


fgdrf:
Thanks for this fix! Looks straight forward, merged it already ;)

PR #254, opened 2017-09-12
FeaturePropertySource fixes and Enhancement of GeomPropertyource

nprigour:
FeaturePropertySource changes:

-  fix certain object casts
- send edit commands only in case of actual value change

GeomPropertySource changes:
- editability improved by providing the ability to view and manipulate coordinates of a geometry in a table. The table can be spawn by pressing ctrl+Alt while selecting or having clicked within the 'well known text field of the FeatureEditor. Functionality is depicted in the image below 
image

### DIFF
--- a/plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/properties/FeaturePropertySource.java
+++ b/plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/properties/FeaturePropertySource.java
@@ -313,21 +313,32 @@ public class FeaturePropertySource implements IPropertySource2 {
             if (id instanceof Integer) {
                 int i = ((Integer) id).intValue();
                 Object attr = feature.getAttribute(i);
+                if (hasValueChanged(value, attr)) {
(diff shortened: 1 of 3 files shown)


fgdrf:
Looks like a new feature to edit Geometry within a table rather then with editing tools. I'm wondering how it works for Multipoligons. Like shown in the screenshout I expected tabled entries grouped by polygon. How the used would know which point is associated to which polygon of the multipolygon ..

@nprigour Did I understood you correctly?

nprigour:
Indeed for multipolygons or multilineString editing of coordinates through this view may be problematic. My intention when I first created this was to have a view where the user can quickly see (and possibly edit) coordinates for a feature since it seems there was no such screen available in udig.
Possibly we can further improve this view in the future. 
My proposal, now that I have a more thorough look, is to pause the integration of this pull request. In the coming days I will try to create  separate  pull requests addressing:

- FeaturePropertySource changes (the ones included here + additional improvements to handle editing and setting of null values)
- editing of geometry coordinates through the table proposed above (and see how we can improve it to address your comments)

PR #257, opened 2017-10-03
fix warnings related to property'propertyId' references

nprigour:
In some plugin.xml files of certain plugins the following warnings appears in eclipse IDE:

> Referenced identifier 'org.locationtech.udig.tool.select.layerHasSelectionProperty' in attribute 'propertyId' cannot be found

or 

> Referenced identifier 'org.locationtech.udig.project.ui.GeometryType' in attribute 'propertyId' cannot be found

Although these warnings do not seem to have an impact in the behavior of the plugins during runtime, the proposed pull request is suggested to clear out these warnings.


fgdrf:
Which version of Eclipse IDE are you using? Do you work with Neon or Oxygen already?

PR #258, opened 2017-11-14
Fix issue causing always full layer export of abstract geometry types

nprigour:
In org.locationtech.udig.catalog.ui.export.CatalogExportWizard when the type of layer is of the abstract type Geometry a selection during the action of a feature export (right click on a layer and selection of  Export->Other->Layer To Shapefile) is ignored causing all layers features to be exported. 


fgdrf:
Is it about exporting only selected features and selection is ignored? Or do you mean the selection within the layers view is ignored and all layers are listed in export dialog? Can you provide a dataset with abstract Geometry type to reproduce the issue. That would help to understand the problem 

Many Thanks

fgdrf:
Absolutely! I could reproduce the bug and test if your fix works for abstract and non abstract geometries. I'm wondering why it's broken for Geometry while it works for e.g. (Multi)Polygons ..

fgdrf:
Great! Works as expected and both, Abstract and Non-Abstract geometries selected features are exported only while unselected are not exported.

Issue #260, opened 2018-01-04
Problem in shapefile saving when copying features from same layer


nprigour:
Hi,
 It seems there is a serious problem when someone attempts to copy from a shapefile layer a feature and paste it to the same layer. A subsequent commit never completes and seems to blocks indefinitely the UI.
This issue was not present in old udig 1.4 (geotools 9.0-M0 and seems to have occured after shifting to locationtech and newer version of geotools (geotools 14.1). 
How to reproduce:
1. select a feature on a shapefile layer and do a right click and select copy from the popup menu
2. click on the map (be sure same layer is selected),  unselect the feature and do again a right click and select paste from the popup menu
3. click the map commit button and observe that the commit progress monitor that appears never completes    
comment: it often occurs that another completely different feature is copied (based on selection and zoom o map after execution of step 2

nprigour:
Hi @fgdrf, 
From a little further investigation it seems that the problem has to do with some read locks being held by another thread from the one that tries to commit (write) to shapefile. More specifically in org.geotools.data.shapefile.files.ShpFiles class (located in gt-shapefile.jar) when method acquireWrite(...) (starting at line 479) is called during commit when a write lock is attempted (see arrow in code provided below) the lock attempt blocks indefinitely.  

```
    /**
     * Acquire a URL for read and write purposes.
     * <p>
     * It is recommended that get*Stream or get*Channel methods are used when reading or writing to
     * the file is desired.
     * </p>
     * 
     * @see #getInputStream(ShpFileType, FileReader)
     * @see #getReadChannel(ShpFileType, FileReader)
     * @see #getWriteChannel(ShpFileType, FileReader)
     * 
     * 
     * @param type the type of the file desired.
     * @param requestor the object that is requesting the URL. The same object must release the lock
     *        and is also used for debugging.
     * @return the URL to the file of the type requested
     */
    public URL acquireWrite(ShpFileType type, FileWriter requestor) {
        URL url = urls.get(type);
        if (url == null) {
            return null;
        }

        // we need to give up all read locks before getting the write one
        Collection<ShpFilesLocker> threadLockers = getCurrentThreadLockers();
        relinquishReadLocks(threadLockers);
    -->    readWriteLock.writeLock().lock();
        threadLockers.add(new ShpFilesLocker(url, requestor));
        mapCache.cleanFileCache(url);
        return url;
    }
```
[...]

PR #261, opened 2018-01-07
fix problem causing deadlock while copying features between the same shapefile layer

nprigour:
fix for problem described in issue #260.

It seems that corresponding feature iterator was never closed for the copied feature collection and the maintained read locks caused problem when trying a write lock was attempted during commit time. 

### DIFF
--- a/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/commands/edit/CopyFeaturesCommand.java
+++ b/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/commands/edit/CopyFeaturesCommand.java
@@ -11,7 +11,6 @@ package org.locationtech.udig.project.internal.commands.edit;
 
 import java.awt.Rectangle;
 import java.io.IOException;
-import java.text.Format;
 import java.text.MessageFormat;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


nprigour:
The present pull request seems to address also [locationtech bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=529519)

fgdrf:
Thanks for providing this patch! Going to test copy & paste within the same layer today ..

plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/commands/edit/CopyFeaturesCommand.java
```diff
@@ -285,7 +288,7 @@ private Query createQuery( ILayer sourceLayer, Filter filter, Layer targetLayer,
 
         queryAttributes.putAll(FeatureUtils.createAttributeMapping(sourceSchema, targetSchema));
         Set<String> properties = new HashSet(queryAttributes.values());
-        return new DefaultQuery(sourceSchema.getName().getLocalPart(), filter, properties
+        return new Query(sourceSchema.getName().getLocalPart(), filter, properties
```
fgdrf: It the switch from DefaultQuery to Query related to the bugfix or just a cleanup here?

nprigour: Just cleanup since DefaultQuery has been marked as deprecated

fgdrf: Tested context menu in map panel with right mouse -> copy (of selected feature) and pased it on map ✔️ 
tested also to copy more than one feature ✔️ 
copy in map-panel and paste onto layer works as expected ✔️ 

Thanks again for fixing this issue!

PR #265, opened 2018-02-07
Fix unit tests

fgdrf:
* configures tycho to run tests against udig-product
* fix tycho issue regarding Bundle-RequiredExecutionEnvironment
* fixed bundle setup regarding bin-ressources
* fixes tests and set a few to @Ignore


fgdrf:
accidently already pushed changes to master ..

PR #270, opened 2018-03-24
splits style.advanced into raster and feature bundle

fgdrf:
Main idea is to create seperate bundles for advanced styling to use these side by side **and** standalone from each other. Asume you have a product that has only support for featuretypes, the customer should not get Raster-Styling pages displayed (the product then would not include *style.advanced.raster* whereas *style.advanced.feature* is included (was *style.advanced*)
 
* cleans codebase regading ColorEditors, removes StolenColorEditor
* moves common usable classes to public packages
* moves common functionality to style.advanced.core module
* renames style.advanced to style.advanced.feature


fgdrf:
Relates to #269

jodygarnett:
Thanks @egouge / @fgdrf - just returning to email now, glad frank was not kept waiting.

PR #272, opened 2018-04-02
remove upper bound restriction during import of com.google.common.base in MANIFEST.MF 

nprigour:
Imports of com.google.common.base package in MANIFEST.MF have a version restriction --> version="[12.0.0,16.0.0)". This prevents upgrade to newer versions of guava library available in latest eclipse orbit releases. 


fgdrf:
Looks good! Please keep in mind, that the version of guave library is pinned in [target-definition](https://github.com/locationtech/udig-platform/blob/master/extras/org.locationtech.udig.target/org.locationtech.udig.target.target#L80) at least for udig-platform build

Issue #256, opened 2017-09-29
consider upgrading udig target env to eclipse neon3 and Java 8


nprigour:
the current latest version of udig uses luna target environment but suffers from a number of limitations/performance issues like:

- drag handles are not available for moving main toolbar and minimized/fast views
- startup time of the udig product seems to be considerably greater compared to old udig 1.4 

upgrading to eclipse neon target environment may resolve this UI issues and improve performance

egouge:
I "upgraded" some parts of uDig to work on a Neon target.  There were not a lot of changes required to get it to work (at least not the parts I am using) and it seems to be working fine.  I cannot comment on overall performance though as I only use a few parts of uDig.

nprigour:
Hi @egouge ,
What was your OS platform? I have also done some preliminary tests of deploying udig to neon target platform and seems to work smoothly in windows (I just updated the target definition to use  neon plugins, set javaSE env to 1.8 in all plugin's  .classpath files and some very minor code modifications in a couple of plugins). 
However testing it in linux I experienced some undesired behavior when for example doing a right click and selecting 'properties' for a layer in the LayersView or a map in the Project Explorer (an extra empty window was spawn without any exception to indicate the reason why).
Besides that the 2 issues I am mentioning in my first comment seems to be resolved

egouge:
Primarily Windows.  I have done a very small amounts of testing on Mac.  I have not tested on linux, but I will try to find time and report back.

fgdrf:
I was trying to upgrade to Oxygen last month while Eclipse released it. As long as I know it was a straight forward upgrade to fix target-platform issues. Going to push that work to forked repository within the next days. 

Questions: Is it prefered to upgrade to Neon first to support Neon (4.6) stable release ([Neon 3](https://wiki.eclipse.org/Neon/Simultaneous_Release_Plan)) while Oxygen (4.7) might have some problems?

nprigour:
I have also tested it a couple of day ago with oxygen (some code modification is further needed due to the fact that some deprecated API was removed). In oxygen both issues mentioned in my initial comments are resolved but still there I experienced the undesired behavior with the empty windows. I will try to upload a screenshot the coming day to show you the problem I am mentioned.

Concerning your question I would say it is better to go to a more mature version of the platform (providing we resolve issues in all supported OS) and this is definetely  Neon.

fgdrf:
[...]

PR #275, opened 2018-04-14
Upgrade eclipse oxygen

fgdrf:
Collaboration with @egouge , @nprigour to upgrade to newer Eclipse release. 

For discussion see issue #256


plugins/org.locationtech.udig.catalog.teradata/src/org/locationtech/udig/catalog/teradata/Activator.java
```diff
@@ -318,7 +316,8 @@ protected Button createButton(Composite parent,
 
 				public File findPluginsDir() throws IOException {
 					Bundle bundle = Platform.getBundle(GEOTOOLS_LIBS_PLUGIN);
-					String filePath = FileLocator.toFileURL(FileLocator.find(bundle, new Path("gtlib"), emptyMap())).getFile();
+					
+					String filePath = FileLocator.toFileURL(FileLocator.find(bundle, new Path("glib"), new HashMap<String, String>())).getFile();
```
nprigour: why the path has changed from "gtlib" to "glib"?

fgdrf: Good question. Just had a look at the complete class and it looks like it doesn't work anyway. The given bundle defined in `GEOTOOLS_LIBS_PLUGIN` doesn't exists anymore, does it?

nprigour: To be honest I have not worked with teradata. The bundle specified by  GEOTOOLS_LIBS_PLUGIN  seems to be absent

fgdrf:
It also turns out, that macosx 32bit isn't a supported platform for SWT anymore.

@nprigour Thanks for this updated regarding orbit installable units. I'll prepare an update!

fgdrf:
Tested local build and if fails with:

> Caused by: java.lang.RuntimeException: Could not determine SWT implementation 
> fragment bundle for environment {osgi.os=macosx, osgi.ws=cocoa, 
> org.eclipse.update.install.features=true, osgi.arch=x86}

It doesn't help to set ``includeAllPlatforms="true"`` ind target definition file..

related Eclipse bugs:
* https://bugs.eclipse.org/bugs/show_bug.cgi?id=491951

fgdrf:
If I disable macosx.cocoa.x88 environment in root pom, the build is fine
```
<!--
<environment>
          <os>macosx</os>
          <ws>cocoa</ws>
          <arch>x86</arch>
</environment>
-->
```
that means we cannot support macosx 32bit environments anymore ..

fgdrf:
Looks like scalebar is broken, it displays wrong scale denominators (at least on my box with German environment). 

image

fgdrf:
The issue with the scale is also on master, so we should fix it independently of this branch. Created issue #276 as bug

fgdrf:
I thought about implications of an upgrade to oxygen. IMO the resulting product needs a JVM >=1.8 installed. For SDK-Users we should/could still support versions lower than 1.8, e.g. 1.7 because our plugins are still compatible for Eclipse Luna, Mars, Neon, and of cause Oxygen. 

This implies that the build would compile on 1.7 level (which is untested becaus of lack of support on Jenkins) 

I‘d like to suggest to enable builds against jdk 1.7 in addition to 1.8 to verify if we are still valid

Opinions?

nprigour:
Hi @fgdrf ,
I do not think that using java 1.7 for building will work with oxygen plugins. There are several  eclipse plugins that have a  Bundle-RequiredExecutionEnvironment: JavaSE-1.8 which menas that java 8 should be used in any case.
Maybe for udig plugins we can use multiple Bundle-RequiredExecutionEnvironment options. I think this is allowed.

nprigour:
Adding to my previous comment, using Bundle-RequiredExecutionEnvironment: JavaSE-1.8 for udig plugins and java 1.8 for building does not prohibit use of  Eclipse Luna, Mars or Neon as target platforms. The Bundle-RequiredExecutionEnvironment parameter specifies the minimum execution environment. Udig plugins may require Java 1.8 (especially if we intend to upgrade to geotools 19.0) but still use eclipse plugins compiled with an older java version. Backward compatibility is guranteed in java.

fgdrf:
Just sorting:
Compile level: Source-code is still valid to compile with 1.7, we don't use JVM features of 1.8++ right now. 

The product we build (against the target file) requires Java 1.8. So it makes sense to think about seperation of concerns: SDK and product. The first should have backward compatibvility which means build with jdk 1.7 and maybe compiled against different target-platformes (mars, neon, oxygen). 
On product level we could build against the latest, Oyxgen at the time of writing, which implies to use JRE 1.8++ to run tests against. If anybody is interested to build against an older Eclipse release we can investigate

Conclusion: If we like to use 1.8 features, we need to set Bundle-RequiredExecutionEnvironment: JavaSE-1.8 otherwise its still valid to use Bundle-RequiredExecutionEnvironment: JavaSE-1.7, isn't it?

Does this makes sense?

nprigour:
IMHO the main issue regarding java build version is related to geotools. If we intend to upgrade to a newer version we have to go for java 1.8 even for source code compilation. Geotools 15.x and onward  are build with java 1.8 therefore I assume we have to go for a source compatibility level of 1.8 even for the SDK build.

nprigour:
I deployed the linux 64bit version in ubuntu 14.04, tried to load a couple of maps, perform some selections and edits, create new featureTypes and its seems to run smoothly without any issues

fgdrf:
@nprigour Great, going to test on mac the build and finally the product..

fgdrf:
Created following piggy back CQ's for Oribit dependencies:
* http://dev.eclipse.org/ipzilla/show_bug.cgi?id=16169 JUnit Version: 4.12 (PB Orbit CQ9298) ✅ 
* http://dev.eclipse.org/ipzilla/show_bug.cgi?id=16170 Apache Batik DOM Level 3 implementation library Version: 1.7.1 (PB Orbit CQ9682) ✅ 
* http://dev.eclipse.org/ipzilla/show_bug.cgi?id=16171 Batik CSS engine library Version: 1.7 (PB Orbit CQ2941) ✅ 
* http://dev.eclipse.org/ipzilla/show_bug.cgi?id=16172 Batik GUI utility library Version: 1.7 (PB Orbit CQ2934) ✅ 
* http://dev.eclipse.org/ipzilla/show_bug.cgi?id=16173 Batik XML utility library Version: 1.7 (PB Orbit CQ2939) ✅ 
* http://dev.eclipse.org/ipzilla/show_bug.cgi?id=16175 Batik utility library Version: 1.7 (PB Orbit CQ2932) ✅ 
* http://dev.eclipse.org/ipzilla/show_bug.cgi?id=16174 Apache Commons Logging Jar Version: 1.1.1 (PB CQ1907) ✅ 
* http://dev.eclipse.org/ipzilla/show_bug.cgi?id=16176 hamcrest-all Version: 1.3 *SUBSET * (PB Orbit CQ7282) ✅ 
* http://dev.eclipse.org/ipzilla/show_bug.cgi?id=16177 hamcrest-core Version: 1.3 (PB Orbit CQ7067)
* http://dev.eclipse.org/ipzilla/show_bug.cgi?id=16178 Miglayout 3.7.1 (PB Orbit CQ3597)
* http://dev.eclipse.org/ipzilla/show_bug.cgi?id=16179 Google Guava Version: 15.0.0 (PB Orbit CQ7768)
* http://dev.eclipse.org/ipzilla/show_bug.cgi?id=16180 itext  2.1.7 Version: 2.1.7 (Subset) (PB CQ3475)
* http://dev.eclipse.org/ipzilla/show_bug.cgi?id=16181 easymock Version: 2.4 (PB Orbit CQ2577)

✅  - approved allready 😃

nprigour:
Hi @fgdrf 
Can you provide some more details regarding this CQ's for Oribit dependencies and what do they imply? it seems that only eclipse committers are able to view them (and I am not one of them :grinning: )

fgdrf:
@nprigour All CQs I filed yesterday are for the dependencies we resolving from orbit repository. In the past we had other versions (e.g. guave 12.0.0 and junit 4.11). 
It was just to get approval from Eclipse foundation and to be prepared for IP review.

A blocker for now is JAI Dependencies we use to build and at runtime which we cannot ship, if we provide a release under Locationtech umprella. However, we can release right now but cannot provide Artifact downloads from any Eclipse Foundation hosting.

Don't get distracted, we move on, merge this branch back and start release train :grinning:

fgdrf:
CI-Job for merged changes to master finished successfully : https://ci.eclipse.org/udig/job/uDig-master-CI/94/ :heavy_check_mark:

PR #290, opened 2018-06-01
match pluginid to file path

egouge:
The point style editor was using the plugin id to find a file path (for Graphics Based Style) and this was causing an error as the plugin id didn't match the file path.  This request updates the pluginid to match the file path.


fgdrf:
Great, thank you for providing this fix!

PR #292, opened 2018-06-26
Disallow DropFilterAction between same layers

nprigour:
DropFilterAction is currently allowed within the same layer. This should be avoided since it may cause some weird result

### DIFF
--- a/plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/dragdrop/DropFilterAction.java
+++ b/plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/dragdrop/DropFilterAction.java
@@ -17,7 +17,7 @@ import org.locationtech.udig.project.command.factory.EditCommandFactory;
 import org.locationtech.udig.project.internal.Layer;
 import org.locationtech.udig.project.ui.internal.Messages;
 import org.locationtech.udig.ui.IDropAction;
-
+import org.apache.commons.lang.ObjectUtils;
 import org.eclipse.core.runtime.IAdaptable;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.geotools.factory.CommonFactoryFinder;
@@ -44,14 +44,17 @@ public class DropFilterAction extends IDropAction {
 
     @Override
     public boolean accept( ) {
-        return true;
+        //allow drop only if destination Layer different than source layer
+        ILayer destinationLayer = getDestinationLayer();
+        ILayer sourceLayer = ((IAdaptable)getData()).getAdapter(ILayer.class);
+        return ObjectUtils.notEqual(destinationLayer, sourceLayer);
     }
 
     @Override
     public void perform( IProgressMonitor monitor ) {
         Layer destinationLayer;
         IMap destinationMap;
-        
+
         if( getDestination() instanceof IMap ){
             destinationMap=(IMap) getDestination();
             destinationLayer=(Layer) destinationMap.getEditManager().getSelectedLayer();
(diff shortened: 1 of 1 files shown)


fgdrf:
@nprigour I guess your change looks reasonable. I'm wondering a bit how to test this change in running application. I'm not aware of this functionality .

Can you help here or point to a help page?

nprigour:
You can check this as follows supposing you have loaded 2 shapefiles to a map. While selection tool is active select a feature i Map editor. Then you can do one of the following actions:

1. hold the right mouse button like doing a drag & drop and release it on the same map editor.
2. hold the right mouse button a do a drag & drop to a layer in Layer view.

Action 1 is the main reason for providing this PR. With the suggested fix it does nothing, while previously it caused a copy of the selected feature on the same layer resulting in duplicate features. This may seemed weird to a potential user since if you were clicking very quickly and selecting features you may accidentally copy features on the same layer.  
Action 2 now results allows copying of features only between different layers and not if the drop is performed on the currently active layer which seems rational.

fgdrf:
I tried the following:
* select an object of a shapefile layer
* selected *this* layer in Layers View
* draged it onto Map View and droped it there

Finally a new layer is listed in Layers View with the same resource

Then I tried to drag the layer on the same layer within Layers View: Nothing happens.

When I use copy and paste ( layer is selected) on Map, then selected  feature is copied within the same ressource (as expected)

I'm wondering how to drag and drop a layer to reproduce behavior you mentioned. I guess I miss something

nprigour:
I did no say to select a layer in layer view or  d&d a layer. What I said is to select a feature in the Map and then while in single slection mode d&d it on the same layer in the Layer view or on the Map itself.

nprigour:
What about this @fgdrf?
Shall we merge and close it?

fgdrf:
> What about this @fgdrf?
> Shall we merge and close it?

I'll try to test it tomorrow again.

fgdrf:
@nprigour Just tested uDig build from current state of master (without your fix) and it works quite well. I was able to copy a feature on the map-panel using Drag&Drop. 

So I'm not sure if your fix is required. Can you test the scenario again? Many Thanks

nprigour:
Yes indeed feature copy is possible I do not deny it. However the reasons for the fix as explained in a previous comment (see my comment on 27 June 2018) are the following (based on my personal experience from an application we have developed):  
1. When working especially with database layers that may have unique constraints such a D&D will cause DB exceptions and rollbacks.
2. More often than not a confused user that works on the map in single selection mode (with a feature already selected) attempts a drag action, assuming he is in multi feature selection mode where drag is perfectly valid. The consequence is that when he realizes his mistake and release the mouse these action results in a drop which causes a copy of the feature which it was not his intention. In the case there is a unique constraint on some layer field this will result in an exce;tion and rollback (possibly invalidating may other actions that may have performed. Even worst if no constraints exist it may result in an undesired feature copy which may not come to his attention. Believe me this has happened numerous times in a cadastral application we have developed based on udig.
3. The accidental copy of feature described above may also occur if you are clicking and selecting features very quickly while in single selection mode   

Based on all the above my humble opinion is that allowing such a D&D within the same layer causes more problems than the ones it solves and I strongly suggest to disable it. If copy between the same layer is desired it can still be achieved by the copy-paste combination of actions which is more or less fully controlled by the user.

fgdrf:
@nprigour Should we merge it or abandon? My read of your last comment is: Let's deactivale D&D of features if source and destination layer are the same. If so, lets prepare another change (for Milestone 2.1.0 ??)

nprigour:
For me its definetely a must merge PR. Maybe we disallow D&D within the same layer but:

1.  we avoid accidental creation of features during selection mode during fast mouse clicks. Try i.e. the following: If you have a polygon layer load it on map, enable 'Feature Selection' tool from the palette and continuously left click the mouse trying to select various polygon features. Verify how easy it is to create a new feature without even noticing it!
2. Copy of features within the same layer is still possible via copy/paste actions so the users is not banned from duplicating a feature. Rather it can do it with full awareness.

PR #293, opened 2018-06-28
fix Feature Table view sorting and selection problems

nprigour:
fix for udig eclipse bug-536390 (https://bugs.eclipse.org/bugs/show_bug.cgi?id=536390) related to TableView behavior during sorting and selection of rows


nprigour:
Hi @fgdrf ,
Shall we also merge this? It is almost  10 months in open status and I do not see any further enhancement to it.

fgdrf:
@nprigour Thanks for this reminder ;) Only thing I'd like you to dO : Please check formatter settings, because whole files are changed and the diff is hard to read. Had origin sources even wrong formattings?

nprigour:
Rebased and forced pushed. Now the actual changes should be evident

PR #294, opened 2018-07-25
Display Charset encoding in LayerSummary of Shapefile layers

nprigour:
Enhancement that displays the character set encoding (only for shapefile resources)  in Layer Summary property page

image


fgdrf:
Just reviewed changeset and I'm wondering about re-formatings. Have you changed Line endings or re-formated whole files?

Good to have a new property to allow translations! Thangs for adding German translastion already :+1:

nprigour:
Hi @fgdrf ,
I really do not know how this bulk reformatting occured. I commit again trying to retain only the changes :)

fgdrf:
Have you configured formatter and automatic save actions. Maybe it auto-format the whole file after saving it rather than only changed blocks.

In the past I tried to keep history where possible and only re-formated blocks where I made any change. What I do not like with this procedure is the different formating within the same ..

Much better your last changes to see what have been modified!

PR #297, opened 2018-09-26
always call EditBlackboard#fireBatchedEvents() in finally block

sloob:
Always call EditBlackboard#fireBatchedEvents() in a try-finally block. This is needed to stop batching events, if an error occurs while the events are batched.

Signed-off-by: sloob <43416137+sloob@users.noreply.github.com>

### DIFF
--- a/plugins/org.locationtech.udig.tool.edit/src/org/locationtech/udig/tools/edit/commands/DefaultCancelEditingCommand.java
+++ b/plugins/org.locationtech.udig.tool.edit/src/org/locationtech/udig/tools/edit/commands/DefaultCancelEditingCommand.java
@@ -60,10 +60,14 @@ public class DefaultCancelEditingCommand extends AbstractCommand implements Undo
 
         EditBlackboard editBlackboard = handler.getEditBlackboard(editLayer);
         editBlackboard.startBatchingEvents();
-        editBlackboard.clear();
-        editBlackboard.fireBatchedEvents();
+        try {
+            editBlackboard.clear();
+        }
+        finally {
+            editBlackboard.fireBatchedEvents();
 
-        handler.repaint();
+            handler.repaint();
+        }
     }
 
     public String getName() {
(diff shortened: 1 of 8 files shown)


nprigour:
Hi @sloob,

Is it possible to explain the nature of problems (possibly by some examples)  that the proposed PR solves in the udig application?
It would be also very helpful if you can provide a way (i.e steps to perform) so that we can demonstrate the problem and the corresponding fix.

sloob:
The EditBlackboard#startBatchingEvents() method increments the batchingEvents counter in the EditBlackboard class. For example, in the DefaultCancelEditingCommand#rollback(IProgressMonitor) method, the call EditBlackboard#startBatchingEvents() is made. Subsequently, various operations are performed (copyFeatures), which can potentially throw an exception. If an exception occurs, the DefaultCancelEditingCommand#rollback(IProgressMonitor) method throws an exception without first calling EditBlackboard#fireBatchedEvents(). Thus, the internal counter is still incremented. All other constellations of startBatchingEvents() and fireBatchedEvents() would never trigger events because the internal counter does not go back to 0.
The fix ensures that the counter is decremented again in case that an exception is thrown.

nprigour:
Hi @sloob ,
Although the proposal to put fireBatchedEvents() method within final blocks seems rational, to be honest I was not able to properly test this situation and verify the exact nature of the problem. I resorted in adding some print statements in the EditBlackboard#startBatchingEvents()  and EditBlackboard#fireBatchedEvents() but what ever editing action I performed, via the UI, I was not able reproduce the internal counter issue you mention. Counter always went back to 0 value. 
Would you be kind enough to propose a combination of Map UI actions that will demonstrate the problem (result in a counter remaining in a value > 0)?

**Note**: It seems that merging of #301 results in a conflict with the present patch.

sloob:
Unfortunately, we have no example constellation where the error occurs. But the point is that the try-finally block makes the program more stable. Or is there anything against merging the PR? I like to resolve the merge conflict, if wanted.

PR #298, opened 2018-09-26
don't select current edit geometry

sloob:
Omit edit geom, cause else we will get a problem clicking in an area of where an edit geom used to be --> the geom is selected again

To replicate the bug: use SelectionTool, then select a geometry in a layer, make the geometry smaller but don’t finalize the geometry. Then do a select in the old area of the geometry. So the old geometry is selected, although there is no more geometry to select.

Signed-off-by: sloob <43416137+sloob@users.noreply.github.com>

### DIFF
--- a/plugins/org.locationtech.udig.tool.edit/src/org/locationtech/udig/tools/edit/commands/SelectFeaturesAtPointCommand.java
+++ b/plugins/org.locationtech.udig.tool.edit/src/org/locationtech/udig/tools/edit/commands/SelectFeaturesAtPointCommand.java
@@ -13,6 +13,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


fgdrf:
triggered a build : https://ci.eclipse.org/udig/job/uDig-PR/29/

nprigour:
Hi @sloob,

To be honest I was not able to replicate the problem the way you describe it. 
The only side effect I observed when executing the steps you mentioned was that the edited feature remained selected even if through the select tool I selected other features and I was not able to deselect it (behavior that still exists after the proposed PR).

sloob:
Hey @nprigour,

You're right. There are constellations where the error does not occur.
First, a shape file for polygon geometry must be loaded, then you draw a new polygon. If you do not save this polygon, then the error described above does not occur. By following the steps described above, the old geometry is not selected when you click in the surface of the old geometry. So the geometry selection works fine.
If the newly drawn polygon is saved, the error described appears. To do this, reduce the geometry on a corner and then click directly into the reduced area. Here, the old, no longer existing geometry is selected again.

nprigour:
Hi @sloob,
I managed to reproduce the problem based on the latest instructions you provided (click directly into the reduced area should occur while in edit mode)

PR #300, opened 2018-09-28
Fixed multiple error-dialogs appearing when a lot of nodes are broken

MSementsov-Ibk:
When selecting multiple nodes and creating an invalid geometry and then pressing enter, multiple error dialogs show up.


nprigour:
Issue confirmed. My only concern is whether it is appropriate to add to EnableAcceptEditCommandHandlerActivator class a dependancy on   org.eclipse.ui.* classes.

akater320:
This is still a problem.
I agree the proposed fix is not as elegant as we might like.
Perhaps we need to get at the root cause of why the acceptCommands are executed 8 - eight - times.
It looks like org.eclipse.e4.ui.bindings.keys.KeyBindingDispatcher.executeCommand(,) is not flagging the event as handled and the event gets fired by every control in the hierarchy as a "traverse" event.

PR #301, opened 2018-09-28
Multi node hanging problem fix

MSementsov-Ibk:
When selecting a bunch of nodes in a dragging them around randomly, some nodes tend to get stuck on other nodes.

### DIFF
--- a/plugins/org.locationtech.udig.tool.edit/src/org/locationtech/udig/tools/edit/support/EditBlackboard.java
+++ b/plugins/org.locationtech.udig.tool.edit/src/org/locationtech/udig/tools/edit/support/EditBlackboard.java
@@ -38,16 +38,26 @@ import com.vividsolutions.jts.geom.Polygon;
 public class EditBlackboard {
 
     protected Map<Point, List<LazyCoord>> coordMapping;
+
     Map<Point, Set<EditGeom>> geomMapping;
+
     private final List<EditGeom> geometries;
+
     private Set<EditBlackboardListener> listeners = Collections
             .synchronizedSet(new HashSet<EditBlackboardListener>());
+
     private Selection selection = new Selection(this);
+
     private volatile int height;
+
     private volatile int width;
+
     private boolean collapseVertices;
+
     PointCoordCalculator pointCoordCalculator;
+
     private volatile int batchingEvents = 0;
+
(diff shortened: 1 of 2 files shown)


nprigour:
Hi @MSementsov-Ibk,

Just to clarify with the term nodes you mean i.e. vertices of a feature of type polygon, multipolygon etc  or just single point features on a point layer?

MSementsov-Ibk:
Hi @nprigour,

yes I do mean vertices, sorry for the confusion.

nprigour:
Can you clarify the following:
- the problem occurs when edit geometry tool is activated?
- the problem may occur when selecting multiple vertices of the same feature or of different features?
- should a certain snapping mode be active (or it is irrelevant)?

I tried to replicate the problem by dragging around some feature vertices while in edit mode but without  luck.

- Do you have an example shp layer (populated with features) that can be used to demonstrate the problem?

MSementsov-Ibk:
Sorry, I'm pretty bad at explaining things, so I made a video demonstrating the issue. 

[current](https://youtu.be/ON3G_ycxEoA)

[fixed](https://youtu.be/AYtWJ5yVYoA)

I hope that helps.

nprigour:
Hi @MSementsov-Ibk 
I have some problems with loading the supplied shp file. Can you add to the zip all the accompanied files (.dbf, .prj, .fix etc.)?

MSementsov-Ibk:
@nprigour 
Sure, here you go. That's all the files I used.
[all_test.zip](https://github.com/locationtech/udig-platform/files/2454760/all_test.zip)

nprigour:
OK issue replicated. It seems that the problem occurs when selecting and moving coordinates of a feature that are very close one to the other.
Although I did not exhaustively reviewed the altered code from a few UI testing on the Map Editor it seems that after applyig the PR moving of features works pretty fine.

Note that If considered for merge code style must be corrected.

MSementsov-Ibk:
Formatting applied.

fgdrf:
Looks like the whole file was re-formated which hides modification for this bugfix a bit. However, thanks for the fix

nprigour:
Since the formatting was applied as a separate commit we can infer the exact changes if we check the PR.
However @MSementsov-Ibk it would be good to apply formatting only to the part of the file you modify. If using eclipse this can be achieved by selecting only the modified fragment of code in the editor and then selecting source-->format from the popup menu  :)

PR #303, opened 2018-09-28
Allow scale ratios of less than 1:1

MSementsov-Ibk:


plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/controls/ScaleRatioLabel.java
```diff
@@ -248,7 +248,10 @@ public void focusLost(FocusEvent e) {
     }
     
     String toLabel( double scaleDenominator ){
-        return "1:" + nf.format( scaleDenominator );
+        if(scaleDenominator >= 1)
+            return "1:" + nf.format( scaleDenominator );
+        else
+            return nf.format( 1.0 / scaleDenominator ) + ":1";
```
fgdrf: I understand this label transformation and wondering about two aspects:
* is it better readable for any user if there is a s scale "2:1" rather than "1:0.5" displayed?
* does the parser can handle it as well to set correct scale if user types e.g. "5:1"?

fgdrf:
Sorry, no response for a year, please re-open pull request if still valid

PR #304, opened 2018-09-28
Repaint Handler so selection box can vanish

MSementsov-Ibk:
When selecting with click&drag the selection box doesn't vanish occasionally.


fgdrf:
Thanks for the fix!

PR #305, opened 2018-10-01
Rollback issue

MSementsov-Ibk:
BehaviorCommand: Commands need to be rolled back from last to first.


fgdrf:
Waiting for Jenkins feedback : https://ci.eclipse.org/udig/job/uDig-PR/31/

plugins/org.locationtech.udig.tool.edit.tests/src/org/locationtech/udig/tools/edit/BehaviourCommandTest.java
```diff
@@ -90,6 +125,8 @@ public Map getMap() {
                 public void run( IProgressMonitor monitor ) throws Exception {
                     if( ran )
                         fail();
+                    Thread.sleep(10);
```
fgdrf: IMO checkes uses a Timestamp rather than a [CountDownLatch](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CountDownLatch.html). 
Otherwise more and more testcases would increase build / test runtime ..

PR #306, opened 2018-10-04
support for text tool contribution and fix of multiple listeners

HendrikPeilke:
We need a text field as a tool contribution in multiple tools inside our applications, so I added support in the ToolOptionContributionItem class. 

Also fixed the code to have no double listeners: addField and fill used both to add a listener to Buttons, so dropped the listener addition in the fill part and let it all be in the addField method, where it was before. (maybe the fill location was better?) At the same time corrected the listen method, where not all listeners were added / removed. 

Signed-off-by: HendrikPeilke <hendrik.peilke@ibykus.de>


nprigour:
I just check and indeed the SelectionListener is called twice. The provided PR fixes these + it adds support for Text Fileds update via the focus listeners. 
Regarding on where exactly to make the listeners addition IMHO and based on the fillFields(...) method description the addField method seems the most appropriate candidate.
I do not see any implication on merging this PR (maybe we just need to remove the commented line 196 to clean up the code).

nprigour:
Hi @fgdrf ,

Maybe we can also include this in 2.1.0?

fgdrf:
@HendrikPeilke Thanks for this contribution! Kodos to @nprigour for testing and review!

PR #307, opened 2018-10-04
externalized strings and added translation into german

sloob:
Externalized some dialog messages and added a translation into german.


plugins/org.locationtech.udig.tools/src/org/locationtech/udig/tools/internal/i18n/messages_de.properties
```diff
@@ -0,0 +1,233 @@
+
```
fgdrf: Just as a hint, We setup a Translation project at Transifex : https://www.transifex.com/udig/udig/dashboard/ where normally resources of origin language (English) synchronized.

However for messages_de.properties I'd suggest o add only keys that are translated already (into German). Here again helps Transifex a lot to get a translation state (in %) any synchronizing translated property files back into codebase.

For this ressource you may have a look at https://www.transifex.com/udig/udig/viewstrings/#de/udig_tools-messages/17409421

fgdrf: If you need help to synchronize German resources, installed Transifex tool installed the Ant-Script [transifex-sync.xml ](https://github.com/locationtech/udig-platform/blob/master/extras/transifex-sync.xml) can help in combination with resource configuration file down below .tx folder : https://github.com/locationtech/udig-platform/blob/master/.tx/config to pull translated message files for a specific languages (or all languages)

fgdrf: @sloob Can we remove German i18n file from this changeset. I'd like to apply this change and merge it into master afterwards.

sloob: Hey @fgdrf, i will look at transifex and if necessary add our translations there. Should I make a new commit without the german i18n file or can you omit the file when merging?

fgdrf: IMO it is quite easy to remove file from commit by using "--amend". This implies to push with "-f" option to rewrite history of your remote branch. I assume nobody else is working on it and therefor it should be a problem.
HTH

fgdrf: Oh, just spotted a "Delete this file" option at web UI menu of this ressource
grafik

I guess this makes it easier than the amend option of git ;)

PR #308, opened 2018-10-05
Efficiency gain by disallowing constant enabling/disabling of the lis…

MSementsov-Ibk:
…tener, otherwise a race condition on the viewPortListener occurs


fgdrf:
Is there a way to re-produce it and if so, how? It would help to test behaviour before and after the change.

MSementsov-Ibk:
`EditBlackboardUtil.disableViewportListener` triggered by deactivating an `EditTool` and 
`EditBlackboardUtil.enableViewportListener` triggered by `EditBlackboardUtil.getEditblackboard` can potentially be called simultaniously by different threads. 

These methods are synchronized, but after the execution of the line 
`listener = new ViewportModelListener(model)`
in `enableViewportListener()` a simultanious call on `disableViewportListener()` (caused by the line `listener = null;`) can cause the line `listener.model.eAdapters().add(listener)` in `enableViewportListener()` to fail.

This fix minimizes the probability of the race condition occuring.

I can't find a method to reproduce this, sorry, it's really occasional. :(

PR #310, opened 2018-10-05
addedCoord instead of toCoord so that undo/redo works properly

MSementsov-Ibk:
The rollback-Method of InsertOnNearestEdgeCommand uses the addedCoord when removing the coordinate so it makes sense that addedCoord is used when adding the coordinate. 


nprigour:
Hi @MSementsov-Ibk ,

Do you have an example indicating a problem that the proposed PR fix?

fgdrf:
Closing due to inactivity, please re-open and provide more details

PR #311, opened 2018-10-05
Allow the setting of a custom ToolManager and MapTracker

MSementsov-Ibk:
Signed-off-by: Michael Sementsov <michael.sementsov@hotmail.de>


plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/ApplicationGIS.java
```diff
@@ -976,9 +988,9 @@ public int getDPI() {
      * @param activeMapTracker the tracker that managers
      */
     public static void setActiveMapTracker( ActiveMapTracker activeMapTrackerToSet ) {
-        if (activeMapTracker != null ){
+        /*if (activeMapTracker != null ){
```
fgdrf: What's the reason behind to comment this block? It there a usecase to allow to re-set map-tracker?

plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/ApplicationGIS.java
```diff
@@ -370,8 +370,20 @@ public static IToolManager getToolManager() {
...
-   }
+    }
 
+    
+    /**
+     * Set the ToolManager singleton.
+     *
+     * @param toolManager the new tool manager
+     */
+    public static void setToolManager(IToolManager toolManager) {
```
fgdrf: Where is this Method called from and what is the reason to (re)set toolManager Singleton?

MSementsov-Ibk:
@fgdrf We use both of these methods inside our own products, where we have a custom ToolManager and ActiveMapTracker. There's no direct benefit for uDig, but having the option to set these would be nice. :)

PR #314, opened 2018-10-10
get message display delay from preference store

sloob:
Adds the possibility to set the delay of displaying messages via the preferences.

Signed-off-by: sloob <sebastian.loob@ibykus.de>

PR #313, opened 2018-10-10
get fill transparency of edit geom from preference store

sloob:
Adds the possibility to specify the factor to reduce the transparency of the color used to fill the geoms on the EditBlackboard.

Signed-off-by: sloob <sebastian.loob@ibykus.de>

PR #315, opened 2018-10-10
get collapse vertices setting from preference store

sloob:
Adds the possibility to change the collapse vertices setting in
EditBlackboard programmatically via the preference store.

### DIFF
--- a/plugins/org.locationtech.udig.tool.edit/src/org/locationtech/udig/tools/edit/preferences/PreferenceConstants.java
+++ b/plugins/org.locationtech.udig.tool.edit/src/org/locationtech/udig/tools/edit/preferences/PreferenceConstants.java
@@ -42,4 +42,9 @@ public class PreferenceConstants {
      */
     public static final int P_DEFAULT_DELETE_SEARCH_SCALEFACTOR = 6;
 
+    /**
+     * The collapse vertices setting in edit blackboard.
+     */
+    public static final String P_COLLAPSE_VERTICES = "P_COLLAPSE_VERTICES"; //$NON-NLS-1$
+
 }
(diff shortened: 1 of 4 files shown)


nprigour:
Hi @sloob, 
Is there a particular reason that the specified property was not included in EditToolPreferences class so that it is available for direct editing in the edit property page UI (similar to what you propose in #314 and #313)?

nprigour:
Hi @sloob, 
Is there a particular reason that the specified property was not included in EditToolPreferences class so that it is available for direct editing in the edit property page UI (similar to what you propose in #314 and #313)?

nprigour:
Hi @sloob, 
Is there a particular reason that the specified property was not included in EditToolPreferences class so that it is available for direct editing in the edit property page UI (similar to what you propose in #314 and #313)?

sloob:
Hi @nprigour,

the problem is that the preference is set in the constructor of the EditBlackboard. If the setting is changed, the existing EditBlackboards do not get that. I also found no way to query all EditBlackboards and set the new preference via the method EditBlackboard#setCollapseVertices(boolean).
Programmatically, the preferences are typically set only once when the application is started, so i guess, the problem described above is not critical here.
If the setting should also be changed via the UI, i suggest the following: With this setting, it should be indicated on the PreferencePage that the change is only active after a restart of the application.
Should i implement the EditToolPreferences in this way or do you have another suggestion?

nprigour:
Hi @sloob 
No no need to change anything in your PR. I was just curious because it was not obvious to me why the properties  in #314 and #313 were exposed to the UI and this one not. Your clarification is sufficient.

fgdrf:
Would please update user-documentation as well that describes what happens on any change of this preference value? In addition due to other changes in this area the current changest has conflict that mus be resolved.

PR #317, opened 2018-10-17
Prevent Exception when the Transform is null

MSementsov-Ibk:


nprigour:
I confirm that this kind of NPE occasionally may occur. In my application I have also added this fragment of code so I consider it pretty useful

PR #318, opened 2018-10-17
Fixed Problems with this method if setGraphics was never called

MSementsov-Ibk:
...and there was no run


fgdrf:
Ate the moment I do not understand how the description of this changeset is related to modified lines. Change indicates that is a safer implemantation which checks for null first before removing listeners. 

Is it possible the not all changes were added to the commit?

Issue #316, opened 2018-10-15
WMS GetFeatureInfo request is malformed


nmtoken:
When using udig (Version: 2.0.0), and trying to do a WMS GetFeatureInfo request, I get the following error:

    <ServiceExceptionReport xmlns="http://www.opengis.net/ogc" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.3.0" xsi:schemaLocation="http://www.opengis.net/ogc http://schemas.opengis.net/wms/1.3.0/exceptions_1_3_0.xsd">
    <ServiceException code="MissingParameterValue">
    msWMSLoadGetMapParams(): WMS server error. Missing required parameter CRS
    </ServiceException>
    </ServiceExceptionReport>

as per below screen shot

![gfi-malfromed](https://user-images.githubusercontent.com/1692891/46964834-0fe86280-d0a1-11e8-84bd-73d83d7f76ce.png)

The actual request sent is (xxx.xxx.xxx substituting for server name):

> http://xxx.xxx.xxx/cgi-bin/UGA_ARGI/ows?SERVICE=WMS&INFO_FORMAT=text/html&LAYERS=ARTISANC&FORMAT=image/png&LANGUAGE=eng&HEIGHT=602&J=256&REQUEST=GetFeatureInfo&I=300&WIDTH=813&BBOX=30.889154617208934,0.03824996332654984,33.29548059763911,1.820055916855415&STYLES=default&QUERY_LAYERS=ARTISANC&VERSION=1.3.0

Adding `&CRS=EPSG:4326&` to the end **AND** switching the BBOX axes to `BBOX=0.03824996332654984,30.889154617208934,1.820055916855415,33.29548059763911&` gives the expected response

fgdrf:
[WMS spec for version 1.3.0](
http://portal.opengeospatial.org/files/?artifact_id=14416) states that CRS is an mandatory parameter for GetFeatureInfo Requests

nmtoken:
Yes indeed,  A GetFeatureInfo request is the effectively the GetMap request (except the the request type is GetFeatureInfo) plus image (pixel) coordinates in the GetMap image for the point clicked on in the map, plus the layers to be queried, plus the information format.  The issue is that the request is automatically generated by udig, so I can't add the CRS parameter manually

nprigour:
Hi @nmtoken ,

[...]

PR #325, opened 2018-10-23
CRS info added in WMS GetFeatureInfo requests

nprigour:
During WMS GetFeatureInfo requests CRS (for WMS version 1.3.0) or SRS (for WMS version 1.1.1) are not provided in the request URL. This may cause problem when retrieving layer information (see also #316 for a more analytical description of the problem)  

### DIFF
--- a/plugins/org.locationtech.udig.info/src/org/locationtech/udig/tool/info/internal/WMSDescribeLayer.java
+++ b/plugins/org.locationtech.udig.info/src/org/locationtech/udig/tool/info/internal/WMSDescribeLayer.java
@@ -202,7 +202,7 @@ public class WMSDescribeLayer {
 
         getmap.setBBox( bbox );
         String srs = CRS.toSRS(bbox.getCoordinateReferenceSystem() );
-        //getmap.setSRS( code != null ? code : srs );
+        getmap.setSRS( code != null ? code : srs );
         
         getmap.setProperty( GetMapRequest.LAYERS, wmslayer.getName() );
         int width = map.getRenderManager().getMapDisplay().getWidth();


fgdrf:
Not really sure about this change, because @jodygarnett fixed getFeatureInfo Request a while ago (https://github.com/locationtech/udig-platform/commit/ec012ac67071f041a29e5cf64664a8e33730abf3) for WMS 1.3.0 requests, where this line has been commented.

@jodygarnett Any ideas why?

nprigour:
A while ago is  almost 6 years back!!. Geotools have undergone several changes since then.
The only thing that the proposed patch does is to set the CRS  parameter (for WMS version 1.3.0 compliant request) or the SRS parameter (for older WMS versions).
Currently this parameter is never sent for GetFeatureInfo requests. All the other information BBOX (in north-east or east-north direction depending on the value of CRS) are  more or less properly set in the request.
I verified behavior against geoserver 2.13 example layers and seems that behavior is correct for both WMS1.3 and WMS 1.1.1 requests. Maybe somebody can check the proposed patch against a MapServer installation to ensure that the problem is addressed

plugins/org.locationtech.udig.info/src/org/locationtech/udig/tool/info/internal/WMSDescribeLayer.java
```diff
@@ -202,7 +202,7 @@ public static LayerPointInfo info2( ILayer layer, ReferencedEnvelope query ) thr
 
         getmap.setBBox( bbox );
         String srs = CRS.toSRS(bbox.getCoordinateReferenceSystem() );
-        //getmap.setSRS( code != null ? code : srs );
+        getmap.setSRS( code != null ? code : srs );
```
fgdrf: Can we cleanup commented line 211 with this pull request?

nprigour: I think yes. I let it there just because it was from the initial committer

nprigour: @fgdrf do you want to remove it?

nprigour:
Hi @fgdrf ,

How should we proceed with this PR;

fgdrf:
Wondering if there is a demo server for mapserver that supports WMS 1.3 out there. I'd like to test against before merging it. I do not expect any issues but lets verify first ..

nprigour:
Try the following:

http://demo.mapserver.org/cgi-bin/wms?SERVICE=WMS&VERSION=1.1.1 (for wms 1.1.1 requests)
http://demo.mapserver.org/cgi-bin/wms?SERVICE=WMS&VERSION=1.3.0 (for wms 1.3.0 requests)
if you then load i.e. the _world cities_ layer and try the info tool (prior to applying the PR) you will get a service exception  indicating:
**WMS server error. Missing required parameter SRS (or CRS)** depending on the version type of the request
If the PR has been applied the getFeatureInfo request returns succesfully and information about the selected city is displayd in infoView.

fgdrf:
Hmm, its working not as stable as expected. I was able to get information for Paris only once. Afterwards no information is shown in Information view. However I only tested with build applictaion and haven't debuged into getFeatureInfo Requests and Responses ..

Any ideas?

nprigour:
Hi @fgdrf,
I do not have any issues retrieving information for the various cities. It works as expected at least  for version 1.1.1 requests. If you check the java console then you will see the requests sent in the following form:
http://demo.mapserver.org/cgi-bin/wms?Y=210&X=512&SERVICE=WMS&INFO_FORMAT=text/html&LAYERS=cities&FORMAT=image/png&HEIGHT=562&REQUEST=GetFeatureInfo&WIDTH=806&BBOX=-26.509857726305366,28.92384602977667,19.297259711417066,60.86379640198511&STYLES=default&SRS=EPSG:4326&QUERY_LAYERS=cities&VERSION=1.1.1 (Note the SRS=EPSG:4326)

Now if used version 1.3 then getInfo requests are also sent in the following form:
http://demo.mapserver.org/cgi-bin/wms?SERVICE=WMS&INFO_FORMAT=text/html&LAYERS=cities&CRS=EPSG:4326&FORMAT=image/png&HEIGHT=562&J=446&REQUEST=GetFeatureInfo&I=246&WIDTH=806&BBOX=-83.43244956259832,-14.396169602825854,-5.934758141128093,39.64068223199584&STYLES=default&QUERY_LAYERS=cities&VERSION=1.3.0  (Note the CRS=EPSG:4326)
However in most cases no information is displayed in info view or if displayed it corresponds to a wrong city. I assume that is because the mapserver is configured to work with version 1.1.1. However this does not prohibit us from verifying correct behavior of sent requests (if CRS or SRS are not provided an error is returned). 
I am testing in windows machine via eclipse IDE.

fgdrf:
I expected a reproducable behaviour for the same city (in this cas Paris). However, I checked with second URL and replaced parameter CRS with SRS or even removed it and server responsed an error. With parameter CRS it doesn't and works as expected

I going to write an unit test to verify CRS is set for GetFeatureInfo Requests if version is 1.3.0

nprigour:
Hi @fgdrf ,

Any progress on this PR?

fgdrf:
> I going to write an unit test to verify CRS is set for GetFeatureInfo Requests if version is 1.3.0

>> Hi @fgdrf ,
>>
>> Any progress on this PR?

I started to write a test but its harder than expected to mock this with EasyMock. Sorry to say but I stoped a while ago (lack of time :()..

Can you help here and provide a test to verify that uDig adds CRS parameter for version 1.3.0 requests...?

PR #326, opened 2018-10-25
[hk] unifies semantic for vertex opacity

fgdrf:


fgdrf:
@nprigour @sloob Thanks for your input. I haven't seen the typo during the review - however, I'll fix it!

Regarding documentation : Sure, an update of [udig-platform/docs/user/en/reference/Edit Tool Preferences.rst](https://github.com/locationtech/udig-platform/blob/master/docs/user/en/reference/Edit%20Tool%20Preferences.rst) makes totally sense. At least the screenshot of https://github.com/locationtech/udig-platform/blob/master/docs/user/en/images/edit_tool_preferences/edittoolpreferences.png

@sloob would you add it, maybe at the end for both pull requests, vertex opacity and fill opacity (#313)?

PR #327, opened 2018-10-26
handle exception during commit or rollback in EditManagerImpl

nprigour:
Proposed PR to provide UI feedback and further actions  when an exception occurs during EditManagerImpl commit or rollback
see also the related discussion at [udig-dev list](https://dev.locationtech.org/mhonarc/lists/udig-dev/msg24414.html)

### DIFF
--- a/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/impl/EditManagerImpl.java
+++ b/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/impl/EditManagerImpl.java
@@ -11,6 +11,10 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArraySet;
 
+import org.locationtech.udig.catalog.CatalogPlugin;
+import org.locationtech.udig.catalog.ICatalog;
+import org.locationtech.udig.catalog.IService;
+import org.locationtech.udig.catalog.IServiceFactory;
 import org.locationtech.udig.core.internal.FeatureUtils;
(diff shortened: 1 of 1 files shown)


sloob:
Hi @nprigour 
thanks for the provided PR.
Maybe it would be better to throw an exception instead of opening a message dialog. So the surrounding application can do whatever it wants with the thrown exception.
For our application, a dialog with an English message is not a good solution.

Thanks!

nprigour:
Hi,

- Regarding the english message indeed you are right. I should try to modify code and internationalize by using the message property files.
- Regarding the message dialog box I assumed this would be a nice solution in order the udig application user to have direct feedback of the error. I think that I experienced with throwing an exception but I had some undesirable results (although this happened some years ago and I did not remember the exact problem). For the udig application opening a message dialog is a good solution however maybe you are right if you use the udig SDK in your custom application. An improvement that I can think  of is to provide a protected method handleCommitRollbackError() in EditMamagerImpl class with a default implementation (that displays the dialog box). Then in your custom application you can easily override  it according to your needs. What do you think of this approach?

fgdrf:
Any progress here for internationalization and what about suggested `handleCommitRollbackError()` method? 

I'm asking myself if its allowed to commit/rollback using API from e.g. some background tasks. If the user requested a rollback using Action in UI I'm fine with a dialog. 
If a process is running, e.g. changes multible features with specific algorithm would EditManager be invokeed to have support for undo/redo as well as transactionsupport? If so I assume a User would interact with the application and needs feedback that something went wrong...

Opinions?

PR #329, opened 2018-11-21
new DrawCoordinate command added

nprigour:
a new DrawCoordinateCommand to UDIG's list of draw commands

### DIFF
--- a/plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/commands/draw/DrawCoordinateCommand.java
+++ b/plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/commands/draw/DrawCoordinateCommand.java
@@ -78,7 +78,7 @@ public class DrawCoordinateCommand extends AbstractDrawCommand {
         } else {
             this.crs = crs;
         }
-        useCircle = true;
+        useCircle = false;
     }
 
     /**
--- a/plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/commands/draw/DrawCoordinateCommand.java
+++ b/plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/commands/draw/DrawCoordinateCommand.java
@@ -1,0 +1,244 @@
+/* uDig - User Friendly Desktop Internet GIS client
+ * https://locationtech.org/projects/technology.udig
+ * (C) 2015, Eclipse Foundation
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * (http://www.eclipse.org/legal/epl-v10.html), and the Eclipse Distribution
+ * License v1.0 (http://www.eclipse.org/org/documents/edl-v10.html).
+*/
+package org.locationtech.udig.project.ui.internal.commands.draw;
+
+import java.awt.Color;
+import java.awt.Paint;
+import java.awt.Point;
+import java.awt.Rectangle;
+import java.awt.Shape;
+import java.awt.geom.Ellipse2D;
+import java.awt.geom.Rectangle2D;
+
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.geotools.geometry.jts.ReferencedEnvelope;
+import org.geotools.referencing.crs.DefaultGeographicCRS;
+import org.locationtech.udig.project.IMap;
+import org.locationtech.udig.project.ui.commands.AbstractDrawCommand;
+import org.locationtech.udig.project.ui.internal.ProjectUIPlugin;
+import org.opengis.geometry.MismatchedDimensionException;
+import org.opengis.referencing.FactoryException;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+import org.opengis.referencing.operation.TransformException;
+
+import com.vividsolutions.jts.geom.Coordinate;
+import com.vividsolutions.jts.geom.Envelope;
+
+/**
+ * Draws a Coordinate at the specified location on the screen viewport.
+ * 
+ * @author nprigour
+ * 
+ * @since 2.0
+ */
+public class DrawCoordinateCommand extends AbstractDrawCommand {
+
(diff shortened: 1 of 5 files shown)


nprigour:
Hi @fgdrf ,
What about this PR? This command may not be used in udig  application at present but it might be useful for potential users of the udig SDK library (you can regard me as a such user )

plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/commands/draw/DrawCoordinateCommand.java
```diff
@@ -0,0 +1,244 @@
...
+                return null;
+            }
+        }
+        return null;
+
+    }
+
+    public void setValid(boolean valid) {
+        super.setValid(valid);
+        if (!valid)
```
fgdrf: Is it necessary to overwride setValid() here? My read of ViewportPainter is that commands get disposed from within `clearCommands`. I guess you can remove method setValid from class because no additional behavior is required here to implement (for example removing listeners).

plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/commands/draw/DrawCoordinateCommand.java
```diff
@@ -0,0 +1,244 @@
...
+            graphics.drawOval((int) ellipse.getMinX(), (int) ellipse.getMinY(),
+                    (int) ellipse.getWidth(), (int) ellipse.getHeight());
+            return;
+        }
+    }
+
+    /**
+     * @return Returns the paint.
+     */
+    public Paint getPaint() {
```
fgdrf: Is this the paintColor for the Coordinate? I guess because internally its a Color object it should return Color here as well. I generall I prefer members to name what it is for, ig. paintColor with named getters and setters (like you almost did for the lineStyle ;))

plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/commands/draw/DrawCoordinateCommand.java
```diff
@@ -0,0 +1,244 @@
...
+        return width;
+    }
+
+    /**
+     * Sets the line style
+     * 
+     * @param lineStyle the style of the line
+     * @param lineWidth the width of the line
+     */
+    public void setStroke(int lineStyle, int lineWidth) {
```
fgdrf: Confusion for me: Here you set two different properties at once. Why not having two different settes for lineStyle (Would an Enum help here btw?) and lineWidth?

Ohh, I guess lineStyle is one of ViewportGraphics.LINE_XXX stroke styles and your implementation is based on DrawShapeCommand or something familiar, don't you?

nprigour: Yes indeed

plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/commands/draw/DrawCoordinateCommand.java
```diff
@@ -0,0 +1,244 @@
...
+                        coord.y + 3, crs).transform(getMap().getViewportModel().getCRS(), true);
+                double[] points = new double[] { bounds.getMinX(), bounds.getMinY(),
+                        bounds.getMaxX(), bounds.getMaxY() };
+                getMap().getViewportModel().worldToScreenTransform().transform(points, 0, points, 0,
+                        2);
+                return new Rectangle((int) points[0], (int) points[1],
+                        (int) Math.abs(points[2] - points[0]),
+                        (int) Math.abs(points[3] - points[1]));
+            } catch (TransformException e) {
+                if (!errorReported) {
```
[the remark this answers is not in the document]
nprigour: Actually as mentioned and in a previous comment this draw command was  based on the DrawFeatureCommand. So I retain the logic adopted there without checking it further.

plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/commands/draw/DrawCoordinateCommand.java
nprigour: Indeed it is not. I will remove it in a following commit

plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/commands/draw/DrawCoordinateCommand.java
nprigour: OK I will fix it.

plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/commands/draw/DrawCoordinateCommand.java
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    /**
+     * @param useCircle the useCircle to set
+     */
+    public void setUseCircle(boolean useCircle) {
+        this.useCircle = useCircle;
+    }
+
+    @Override
```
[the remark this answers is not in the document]
nprigour: Yes there is no test case as is the case with all DrawCommands. If you think it absolutely necessary, maybe I can provide a test in the form of a test menu item that draws a circle at a coordinate on a map. But it could be tested manually and not be automated. What do you think?

nprigour:
Hi @fgdrf ,

I added a test handler which can be used for testing DrawCoordinateCommand via the UI. The handler centers viewport in the input coordinate and  draws a flashing dot for a couple os seconds (see screenshot below)
εικόνα

nprigour:
Hi @fgdrf ,

what about this PR? It's been quite some time since the test handler was added. Shouldn't we merge  and close it?

PR #331, opened 2018-12-18
use default CRS if it's set in preference store

sloob:
Fix bug, that the default CRS is never used, although a valid and not
"bad default" CRS is specified in the preference store.


nprigour:
This PR should me marked as duplicate of #352.

fgdrf:
@sloob We changed codebase recently and idendified your change is simmiliar to #352

PR #332, opened 2018-12-27
fix possible NPE exception due to null CRS during SelectAllFearureOp execution 

nprigour:
if during execution of SelectAllFearureOp, CRS is not available at layer level then a NPE exception is thrown and features are not selected. In that case, the suggested fix attempts to obtain CRS from the Map (through the ViewportModel) so that selection is possible

### DIFF
--- a/plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/operations/SelectAllFeaturesOp.java
+++ b/plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/operations/SelectAllFeaturesOp.java
@@ -56,6 +56,9 @@ public class SelectAllFeaturesOp implements IOp {
         ReferencedEnvelope bounds = viewportModel.getBounds();
         CoordinateReferenceSystem dataCrs = schema.getCoordinateReferenceSystem();
         
+        if (dataCrs == null) {
+        	dataCrs = viewportModel.getCRS();
+        }
         ReferencedEnvelope newBounds = bounds.transform(dataCrs, true);
         
         String name = geometryDescriptor.getLocalName();


nprigour:
Hi @fgdrf ,

what about this PR? THe fix is quite obvious. I do not see any reason keeping us from merging it :)

fgdrf:
@nprigour Agree, your change looks good and I do not expect any problems. I'm wondering what a user would expect as a result. My read: The current BBox of the ViewPortModel is used to define a bbox filter which is not "All" in terms the function from name pretend.

@moovida Can you help here & describe the usecase behind this function

nprigour:
THe purpose of this PR is to allow selection of features in a layer even if  no CRS is found. An example use case where this has appeared is when loading in a Map with predefined projection a layer with a missing .prj file. In that case the layer CRS is null and if you try the operation --> select All features  (in order i.e. later on to copy paste them in another layer) from the layersView popup the action will fail (no features will be selected). The present fix makes this selection possible .
This is often the case in our application where we are given shp files without the .prj file and we want to copy paste them to a DB layer.

fgdrf:
Yeah, got it. However I'm still confused because of the bbox filter which result in a collection of features within the current shown and visible area whereas the source might have more features in the store outside of the current rectangle/bbox. if the function would be name like "Select all features in visible area" or something similiar.

I understand your fix and this is fine! I just have a different expectation due to the semantic of "SelectAllFearureOp". I'd prefer to use Filter.INCLUDE to get All features indpendenly from the current viewport bounding box and CRS. This might result into refactoring the whole class and remove BBoxFilter-Implementation...

nprigour:
Ok your comment may be valid however this may be addressed by another pull request. In general selection of features works on the viewport who's is more or less the default active or processing region. 
We may also investigate the possibility to rename the operation to something like allFeaturesInView

fgdrf:
The class I had in mind would look like that (name represents the operation):

```
public class SelectAllFeaturesOp implements IOp {

    public void op( final Display display, Object target, IProgressMonitor monitor ) throws Exception {

        final ILayer layer = (ILayer) target;
        
        SelectionCommandFactory cmdFactory = SelectionCommandFactory.getInstance();
        UndoableMapCommand selectCommand = cmdFactory.createSelectCommand(layer, Filter.INCLUDE);
        
        IToolContext toolContext = ApplicationGIS.createContext(ApplicationGIS.getActiveMap());
        toolContext.sendASyncCommand(selectCommand);
    }
}
```

see `Filter.INCLUDE`

Opinions?

nprigour:
I have no objection to add the suggested operation. However I strongly insist on keeping the existing SelectAllFeaturesOp even if you think we should rename it to something more specific according to its actual behavior (i.e SelectAllFeaturesInViewPort).
That is because IMHO selecting features in a view is most of the cases what we actually want. I.e. for JDBC datastores containing hundreds of thousands or even million of features I usually want to select what I see in the viewport in order to copy or move it to another layer rather than select everything.

PR #333, opened 2019-02-14
improve DistanceTool to work correctly during Map zoom in and zoom out actions

nprigour:
When pwrforming a zoom in or zoom out while using the DistanceTool the DistanceTool lines do not scale correctly (actually they do not scale at all) together with the map leading to weird results. This PR attempts to fix this by storing clicked points as coordinates and not points and performing the conversion during final rendering on screen.
Another enhancement introduced by this PR is that the displayed information at the bottom left of the screen includes also information about the last segment length. 


fgdrf:
Great, thanks for contributing. I'll test it and give you feedback. In between I triggered Jenkins to give feedback first : https://ci.eclipse.org/udig/job/uDig-PR/39/

nprigour:
The jenkins failure is irrelevant to the present PR. However I think we had this kind of failure in the past around one year ago and somehow it was resolved. Now it seems to appear again. Strange things!

plugins/org.locationtech.udig.tool.measure/src/org/locationtech/udig/tool/measure/internal/messages_ru.properties
```diff
@@ -1,2 +1,2 @@
 DistanceTool_error=\u041d\u0435\u0432\u043e\u0437\u043c\u043e\u0436\u043d\u043e \u0432\u044b\u0447\u0438\u0441\u043b\u0438\u0442\u044c \u0440\u0430\u0441\u0441\u0442\u043e\u044f\u043d\u0438\u0435
-DistanceTool_distance=\u0420\u0430\u0441\u0441\u0442\u043e\u044f\u043d\u0438\u0435 \= {0}
\ No newline at end of file
+DistanceTool_distance=\u0420\u0430\u0441\u0441\u0442\u043e\u044f\u043d\u0438\u0435 \= {0} (\u043F\u043E\u0441\u043B\u0435\u0434\u043D\u0438\u0439 \u0441\u0435\u0433\u043C\u0435\u043D\u0442 \= {1})
```
fgdrf: Cannot check this ;)

fgdrf: Tested new functionality and it works as expected. Awesome to allo scrolling/zooming while tools is still active.

plugins/org.locationtech.udig.tool.measure/src/org/locationtech/udig/tool/measure/internal/messages_de.properties
```diff
@@ -1,2 +1,2 @@
-DistanceTool_distance = Entfernung \= {0}
+DistanceTool_distance = Entfernung \= {0} (letztes segment \= {1})
```
fgdrf: What do you think about "Gesamtstrecke \= {0} (aktuelles Segment \= {1})"

fgdrf:
checked current docs and I guess the page is still valid : http://udig.github.io/docs/user/reference/Information%20Tools#distance-tool

plugins/org.locationtech.udig.tool.measure/src/org/locationtech/udig/tool/measure/internal/messages_de.properties
nprigour: OK modification commited after all you are the German guy :)

plugins/org.locationtech.udig.tool.measure/src/org/locationtech/udig/tool/measure/internal/messages_ru.properties
nprigour: > 
> 
> Cannot check this ;)

My either I just rely in google translation!!

nprigour:
For the italian translation from the little I know we are OK.
Now about the rest translations, the initial versions of the message files contained an entry for the  _DistanceTool_distance_  property which I just augment it with the string after the 1st argument. If I restore it as it was before then we may have a problem since the MessageFormat.format(...) call in line 247 expects 2 extra arguments. Unless we decide to replace the strig within the parenthesis with the english translation?

fgdrf:
I'd prefer to remove properties files we not totally sure about translation. Would you remove all oher message files than messages.properties, messages_de.properties, and messages_it.properties (if you are sure about)?

fgdrf:
I'd like to suggest a little optimization. What do you think about switching the display of the last segement and total distance such as "Distance: {0} (total: {1})" **and** displaying the information directly at the cursor rather than in the Status bar:

here is a little example:

grafik

I can create a pull request for your branch...

nprigour:
Why not have them both? In any case you can proceed with the suggested PR in order to check behavior

fgdrf:
> Why not have them both?

Do you mean having a rendered label at cursur in Map and updted statusbar?

> In any case you can proceed with the suggested PR in order to check behavior

✔️  Done, its fine from my perspective and ready to merge except i18n ressources we should remove from this pull ;)

> By the way will you provide an indication of length for each segment as seen on map?

Yeah, forgot to mentioned it, every segment could get rendered a label as shown. However it might have conflicts to other labels of other segments. Could be optional ..

nprigour:
Hi @fgdrf ,
Yes we can keep the  status bar info and you can also add the information on the map as you suggest in the screenshot.

Now I thought you will provide a PR on my existing PR branch. Isn't that the case? Do you want to proceed first with merging the present and you create a separate PR on the master branch?. If yes, please confirm and I will proceed with the removal of the i18n resources.

Now regarding  the labelling in each segment indeed it may cause some clutter but in many cases will be very useful. May be you can control it by an appropriate preference option?

fgdrf:
@nprigour Sorry for confusion, lets proceed with this PR as it is and I'll create another with minor optimizations. It takes a bit longer :P

For now, lets proceed with the removal of the i18n resources and we can merge this PR! Thanks in advance for fixing it!

Preferences Option is a good point I'll address!

nprigour:
OK remove i18n for es, eu, ru, ko country codes

PR #337, opened 2019-03-10
Enable emition of ToolLifecycleEvent during activation/deactivation of a Modal Tool

nprigour:


nprigour:
I do not know why emition of  lifecycle events during tool activation/de-activation is not there while it is present when setting enabled status!

PR #338, opened 2019-03-13
Fix exceptions thrown by LineTool and PolygonTool

akater320:
With "Advanced Editing" enabled in Preferences, PolygonTool and LineTool will throw many exceptions when dragging existing vertices or dragging the entire geometry.

AdvancedFeaturesEventBehavior attempts to composite its sub-behaviours. SelectVerticesWithBoxBehaviour cannot be successfully composited with the move behaviours since they both require the EditToolHandler lock.

The selection command is added to the undo stack, which is consistent with the Edit Geometry (SelectionTool) tool.

### DIFF
--- a/plugins/org.locationtech.udig.tool.edit/src/org/locationtech/udig/tools/edit/impl/LineTool.java
+++ b/plugins/org.locationtech.udig.tool.edit/src/org/locationtech/udig/tools/edit/impl/LineTool.java
@@ -124,10 +124,12 @@ public class LineTool extends AbstractEditTool {
         helper.add( new MoveVertexBehaviour() );
         helper.add( new MoveGeometryBehaviour());
         helper.stopMutualExclusiveList();
+        helper.stopAdvancedFeatures();
         
-
+        helper.startAdvancedFeatures();
         helper.add( new SelectVerticesWithBoxBehaviour() );
         helper.stopAdvancedFeatures();
+        
         helper.add( new AcceptOnDoubleClickBehaviour() );
         helper.add( new SetSnapSizeBehaviour());
         helper.add( new StartExtendLineBehaviour() );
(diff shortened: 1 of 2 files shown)


nprigour:
Hi @akater320,

Can you become a little more specifc on the kind of exceptions thrown and provide some info on the conditions and steps under which we can reproduce them. I tried to perform some dragging of vertices (even of the whole geometry) in a polygon jdbc layer  and I did not experienced any such issue.

akater320:
Thanks for checking it out. 
You're right that the run-time behavior appears correct. The main issue is that performance/frame-rates are being degraded by using exceptions for control flow. Also, the console and logs are spammed by stack traces, making debugging difficult. The attached log file covers about 0.5 seconds of using the polygon tool with advanced features turned on.
[udig-error-log.txt](https://github.com/locationtech/udig-platform/files/2978041/udig-error-log.txt)

nprigour:
These errors are logged  every time you try and drag vertices? Is it possible that it is some issue related to Mac OS only? I cannot verify them problem since I work on windows env (windows 10) and I have never seen these stack traces while working in advanced editing mode. 
Is any body else using udig in Mac env that can offer some help here?

akater320:
To reproduce, on Windows 10 or MacOS:
1. Layer > Create (Polygon is easiest to demonstrate.)
2. Activate Polygon Tool
3. Ensure Advanced features are enabled for the tool
3. Select vertices with the box.
4. Drag vertices
5. Exceptions are thrown on every mouse move event.

https://drive.google.com/open?id=1Rs6azz_kau-K6mv_EEbv4g_-B5IbeV9K

nprigour:
OK thanks @akater320 issue verified/replicated based on your instructions.
Based on some UI testing performed, it seems that the provided changes fix the problem

nprigour:
Hi @fgdrf ,
Do we want something more related to this issue? From the manual testing performed, it seems that the problem is fixed and I do not see any implications by the added code.

Issue #342, opened 2019-03-20
Hole Cutter tool cannot implicitly close triangular holes


akater320:
To reproduce:
1. Activate the hole cutter tool.
2. Begin a hole by placing 2 vertices within a polygon.
3. Observe the status bar message "...double-click to end."
4. Attempt to complete the hole by double clicking. 
- The hole is not finished. 
- An exception is logged.
- A third vertex is placed.
5. Now with the 3 vertices, attempt to finish the hole by pressing return.
- An exception is logged instead.

(A related problem is that a triangular polygon drawn with CCW winding will not be converted to CW. You'll need a debugger to observe this.)

PR #343, opened 2019-03-20
Calculate winding from ring.

akater320:
Correctly calculate the winding of triangles.

Fixes [342](https://github.com/locationtech/udig-platform/issues/342)

Signed-off-by: Adam Kater <adam.kater@gmail.com>

### DIFF
--- a/plugins/org.locationtech.udig.tool.edit/src/org/locationtech/udig/tools/edit/support/GeometryCreationUtil.java
+++ b/plugins/org.locationtech.udig.tool.edit/src/org/locationtech/udig/tools/edit/support/GeometryCreationUtil.java
@@ -232,7 +232,7 @@ public class GeometryCreationUtil {
         LinearRing shell = GeometryBuilder.create().safeCreateLinearRing(shellCoords);
 
         try {
-            if (CGAlgorithms.isCCW(shellCoords)) {
+            if (CGAlgorithms.isCCW(shell.getCoordinates())) {
                 shell = JTSUtilities.reverseRing(shell);
             }
         } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


nprigour:
Indeed the suggested commit fixes the issue mention in #342 and permits creation of holes with 3 coordinates and suppress the thrown exception.
However @akater320 would it be possible to explain a little bit the issue with the winding? Debugging I can infer that all holes are converted to CW (ClockWise) winding. Do you think that there is a particular reason for doing this or is it just an arbitrary assumption by the class creator in order all holes to have the same winding? Would there be any implication that we must be aware of?

akater320:
Good question.
I think the author's intent was to orient the edges so that the interior of the polygon is always on the right.

But, assuming this is the correct Polygon class [com.vividsolutions.jts.geom.Polygon](https://github.com/simplegeo/jts/blob/master/src/com/vividsolutions/jts/geom/Polygon.java) the comments explicitly state that orientation of rings does not matter. There is also no guarantee that orientation is preserved.
Java 2D uses an even/odd ray-cast to determine the interior of closed Paths so orientation won't matter there either.

So, the winding direction should probably be treated as arbitrary.

nprigour:
OK thanks for the clarification @akater320.
Now in order to finalize the issue I think (please @fgdrf  confirm) that we should add a test that verifies that correct behavior of GeometryCreationUtil when creating polygon with a hole with 3 coordinates. There is a test class GeometryCreationUtilTest in plugin org.locationtech.udig.tool.edit.tests which can be extended to include a test on that. Would it be possible to do it?

akater320:
@nprigour Is it possible to run all of the tests at once? Or is it necessary to run each test plugin individually?

nprigour:
What do you mean by all the tests at once? If you refer to all udig tests in the various plugins then this can be accomplished by mvn by running the following command 
mvn clean install -Pproduct,test

akater320:
@nprigour 
Perfect! Thank you.

fgdrf:
Just testing a maven run with test profile, if I get no failures I'll merge it

Issue #346, opened 2019-04-04
Error dialog displayed when creating layer.


akater320:
An error dialog is displayed when creating a layer if the last used tool was from the "Create" category.

To reproduce:
1. Load a vector layer in the editor. e.g. Layer > Create
2. Active any of the "Create ..." tools. e.g. "Create Line"
3. Close all editor tabs.
4. Create a new vector layer. i.e. Layer > Create 
--- (DO NOT use the default layer name. This is likely to cause an NPE if another layer already has the same name.)
5. An error dialog is displayed.

I think the dialog is erroneous. The new layer is created correctly and the edit tool works as expected.

PR #341, opened 2019-03-19
Fix for zoom/pan flicker issue

akater320:
Stale buffers are now transformed to the current viewport before rendering.
Fixes [#334](https://github.com/locationtech/udig-platform/issues/334)

PR #347, opened 2019-04-04
Return a safe default rather than null.

akater320:
The interface is not clear whether null is allowed to be returned here. However, no clients currently test for null. This now returns a safe default instead.
Signed-off-by: Adam Kater <adam.kater@rgi-corp.com>

### DIFF
--- a/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/ViewportModelImpl.java
+++ b/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/ViewportModelImpl.java
@@ -685,7 +685,7 @@ public class ViewportModelImpl extends EObjectImpl implements ViewportModel {
      */
     public AffineTransform worldToScreenTransform() {
         if (!validState())
-            return null;
+            return new AffineTransform(); //Identity (screen-space is arbitrary here.)
         // set up the affine transform and calculate scale values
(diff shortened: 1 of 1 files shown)


akater320:
Fixes [346](https://github.com/locationtech/udig-platform/issues/346)

Prevents NullPointerExceptions when an edit tool is deactivated or activated while its map is closed.

plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/ViewportModelImpl.java
```diff
@@ -685,7 +685,7 @@ public void setBounds(double minx, double maxx, double miny, double maxy) {
      */
     public AffineTransform worldToScreenTransform() {
         if (!validState())
-            return null;
+            return new AffineTransform(); //Identity (screen-space is arbitrary here.)
```
fgdrf: Aha, now I got the puzzle and it looks like this change addresses requested Null-check in pull-request #341 ([review discussion](https://github.com/locationtech/udig-platform/pull/341#discussion_r278718919))

fgdrf: It would be great to have another test for worldToScreenTransform for both cases, validateState is true or false ;)

Issue #351, opened 2019-08-08
Wrong behavior in InitMapCRS if defaultCRSPreference is set


fgdrf:
Implementation in InitMapCRS has wrong behavior. If a product is configured using e.g. Pseudo-Mercator with:
`org.locationtech.udig.project/defaultCRSPreference=3857`

the first added layer changes ViewportModel CRS to its given (different) CRS (e.g. EPSG:4326)

Expected behavior: Whenever a different default is configured and EPSG code is fine, the ViewportModel CRS should be unchanged whenever a layer is added to the map.

see second part of if-statement:
https://github.com/locationtech/udig-platform/blob/38712e7275db02e3961c30cedbbc782658351693/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/impl/InitMapCRS.java#L42

PR #352, opened 2019-08-22
[#351] fixes InitMapCRS Interceptor

fgdrf:
If defaultCRSPreference is set and valid, do not use the CRS from first added layer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,27 @@
 sudo: true
+dist: trusty
 language: java
 jdk: oraclejdk8
-
-python:
-  - "2.7"
-
-env:
-  global:
-    - TRAVIS_PYTHON_VERSION="2.7"
-    - MVN_VERION="3.3.9"
-
 cache:
   directories:
   - $HOME/.m2
+python:
+  - "2.7"
+env:
+  global:
+    - TRAVIS_PYTHON_VERSION="2.7"
+    - MVN_VERION="3.5.4"
+    - MAVEN_OPTS="-Xmx8g -XX:-UseGCOverheadLimit"
 
 before_install:
-  - echo "MAVEN_OPTS='-Xmx8g -XX:-UseGCOverheadLimit'" > ~/.mavenrc
+  - rm ~/.m2/settings.xml
+  - unset _JAVA_OPTIONS
+
+before_script:
+  - wget https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/${MVN_VERION}/apache-maven-%{MVN_VERION}-bin.tar.gz
+  - tar zxf apache-maven-${MVN_VERION}-bin.tar.gz
+  - export M2_HOME=$PWD/apache-maven-${MVN_VERION}
+  - export PATH=$M2_HOME/bin:$PATH
 
 install:
   - "sudo pip install 'requests[security]'"
@@ -30,8 +36,3 @@ notifications:
   email:
     on_success: never
     on_failure: never
-
-branches:
-  only:
-    - master
-    - /^travis.*$/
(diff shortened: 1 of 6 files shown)


fgdrf:
@nprigour Would you please have a look at the Interceptor changes? The problem is described in issue #351

nprigour:
Hi @fgdrf ,

First of all I would like to ask why you have collocated travis build changes within the present request? Are they relevant? 

Now about the InitMapCRS Interceptor:
I review and test via the UI the proposed changes. Indeed now the behavior is as follows:

-  if -1 is set then the first layer determines the map CRS
- if a specific EPSG code is set (i.e. 4326) and is valid then the map is created with that CRS independent of the 1st added layer CRS. 
However there is still an issue:
The defaultCRSPreference  set via the Map preference  page is somewhat cached  the first time (since it is set in the field DEFAULT_CRS of the ViewportModel interface). Therefore any subsequent change via the preference page is not propagated to the ViewportModelImpl and the DEFAULT_CRS remains always to the value set the 1st time. You can verify this by changing, while udig is running, the CRS preference value to another EPSG code =2100, deleting the Map and try to load a shapefile to the LayersView. You will observe that the Map CRS will remain 4326.
A possible solution would be to not use the value DEFAULT_CRS when setting the cRS protected member in ViewportModelImpl class (as in line 91): 

       `protected CoordinateReferenceSystem cRS = DEFAULT_CRS;`

but rather modify it to something like the following:
        
        `protected CoordinateReferenceSystem cRS = ViewportModelImpl.getDefaultCRS();`

However we must also take care of other fields that dependent on DEFAULT_CRS (i.e. from a quick check there is also NIL_BBOX which is created and cached based on DEFAULT_CRS)

fgdrf:
> First of all I would like to ask why you have collocated travis build changes within the present request? Are they relevant?

Not really but Travis is the only system we have to get feedback regarding compile / unit-test failures. Pull-Request build on ci.eclipse.org doens't work anymore..

> Now about the InitMapCRS Interceptor:
> I review and test via the UI the proposed changes. Indeed now the behavior is as follows:
> 
>     * if -1 is set then the first layer determines the map CRS
> 
>     * if a specific EPSG code is set (i.e. 4326) and is valid then the map is created with that CRS independent of the 1st added layer CRS.

Great, thanks for testing!

>       However there is still an issue:
>       The defaultCRSPreference  set via the Map preference  page is somewhat cached  the first time (since it is set in the field DEFAULT_CRS of the ViewportModel interface). Therefore any subsequent change via the preference page is not propagated to the ViewportModelImpl and the DEFAULT_CRS remains always to the value set the 1st time. You can verify this by changing, while udig is running, the CRS preference value to another EPSG code =2100, deleting the Map and try to load a shapefile to the LayersView. You will observe that the Map CRS will remain 4326.
>
>       A possible solution would be to not use the value DEFAULT_CRS when setting the cRS protected member in ViewportModelImpl class (as in line 91):
>       ```
>        `protected CoordinateReferenceSystem cRS = DEFAULT_CRS;`
>       ```
> but rather modify it to something like the following:
> 
> ```
>     `protected CoordinateReferenceSystem cRS = ViewportModelImpl.getDefaultCRS();`
> ```
> 
> However we must also take care of other fields that dependent on DEFAULT_CRS (i.e. from a quick check there is also NIL_BBOX which is created and cached based on DEFAULT_CRS)

Sounds like another issue independently from this changeset for LayerInterceptor. Would you help providing a fix for this problem to referenced branch as a pull request? So we could join forces to fix both problems ;)

nprigour:
Hi @fgdrf ,

I added a PR in your local fgdrf:ISSUE_351_fix  branch with a proposed solution

Issue #244, opened 2017-07-07
No appenders could be found for logger


fgdrf:
```
og4j:WARN No appenders could be found for logger (org.geotools.factory).
log4j:WARN Please initialize the log4j system properly.
```

This is part of a general problem of a number of warning messages being issued as various open source projects start up.
The following clarification was offered on the user list

Hi Donovan,
This is related to the way that log4j uses reflection to find the appender.
I haven't looked deeply within the new udig 1.3.0 code base yet so I am not sure where things have moved and indeed if there is a log appender, but I did notice that there is a bundled org.apache.log4 library.
I have been running with SLF and log4j logging in my RCP plugins for some time and there are a couple of things that need to be done to permit log4j to work from plugins, please view the notes below:
1. The bundled org.apache.log4j library provided in the udig code base doesn't have the correct directive in it to enable log4j to work, you need an Eclipse-BuddyPolicy: registered directive in the manifest of org.apache.log4j as follows:
```
Manifest-Version: 1.0
Bundle-ManifestVersion: 2
Bundle-Name: org.apache.log4j
Bundle-SymbolicName: org.apache.log4j
Bundle-Version: 9.0.1.20110423
Bundle-Vendor: Apache Log4j
Export-Package: org.apache.log4j,
org.apache.log4j.chainsaw,
org.apache.log4j.config,
org.apache.log4j.helpers,
org.apache.log4j.jdbc,
org.apache.log4j.jmx,
org.apache.log4j.lf5,
org.apache.log4j.lf5.config,
org.apache.log4j.lf5.util,
org.apache.log4j.lf5.viewer,
org.apache.log4j.lf5.viewer.categoryexplorer,
org.apache.log4j.lf5.viewer.configure,
org.apache.log4j.lf5.viewer.images,
org.apache.log4j.net,
org.apache.log4j.nt,
org.apache.log4j.or,
org.apache.log4j.or.jms,
org.apache.log4j.or.sax,
org.apache.log4j.spi,
org.apache.log4j.varia,
org.apache.log4j.xml
Eclipse-BuddyPolicy: registered
Bundle-RequiredExecutionEnvironment: JavaSE-1.6
```
2. You will need to place a PluginLogAppender class in one or more plugins - see http://www.eclipsezone.com/eclipse/forums/t55300.html - http://www.eclipsezone.com/eclipse/forums/t55300.html. The example I use follows:
```
/*******************************************************************************
Copyright (c) 2005 John J. Franey
[...]

PR #360, opened 2019-10-18
add a log4j appender

egouge:
See issue: https://github.com/locationtech/udig-platform/issues/244


plugins/org.locationtech.udig.libs/log4j.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
fgdrf, line 1: I guess its required to add log4j.xml to build.properties (bin.includes) otherwise it might not be present in final jar.

plugins/org.locationtech.udig.libs/META-INF/MANIFEST.MF
```diff
@@ -8,6 +8,7 @@ Bundle-Localization: plugin
 Eclipse-BundleShape: dir
 Eclipse-BuddyPolicy: registered,
  ext
+Eclipse-RegisterBuddy: org.apache.log4j
```
fgdrf, line 11: Would it help to define another bundle to separate concerns, in this case logging with log4j to e.g. org.locationtech.udig.log4j ? So it might be easier for potential developers to use another logging framework rather than log4j. 
Other could contribute a commons.logging adapter or even a slf4j logging adapter as well.

plugins/org.locationtech.udig.libs/src/org/locationtech/udig/libs/internal/PluginLogAppender.java
```diff
@@ -0,0 +1,132 @@
+/*******************************************************************************
+Copyright (c) 2005 John J. Franey
```
fgdrf, line 2: Do we need to raise a CQ request at Eclipse Foundation for this copied code fragments? I'm not sure about the procedure to integrate source code from other projects ..

plugins/org.locationtech.udig.libs/log4j.xml
egouge: Added.

plugins/org.locationtech.udig.libs/META-INF/MANIFEST.MF
egouge: I feel like if somebody wants to do this they can. I'm not sure there is a reason for us to do it now.

plugins/org.locationtech.udig.libs/src/org/locationtech/udig/libs/internal/PluginLogAppender.java
egouge: I also have no idea how this works.  I thought because it was EPL it was ok, but maybe not.

plugins/org.locationtech.udig.libs/META-INF/MANIFEST.MF
fgdrf: I had a look at it, please see pull request https://github.com/egouge/udig-platform/pull/7

@egouge much appreciate if you could have a look at and test in your environment (SMART??)

egouge: I merged this and it seems to work as expected

fgdrf:
Just a minor issue reported by Travis : the version of the module is not
correct anymore and doesn’t fit to parents version. Could you please change
MF to 2.3.0.qualifier and in Pom to 2.3.0-SNAPSHOT

I’ve increase version Number on master in between while 2.2.0RC1 is out



Emily <notifications@github.com> schrieb am Do. 2. Juli 2020 um 21:06:

fgdrf:
@egouge Thanks a lot for contribution and testing!

Issue #363, opened 2019-12-20
Memory Leak of FeatureEvents in Collection and performance issue


fgdrf:
If running uDig of a long time with a lots of feature events for different layers, the collection increases constantly of time. There is a clean() call in add(index, object) that clear the cache of amount of features it larger than 10.

https://github.com/locationtech/udig-platform/blob/c0990a643d91af69038e2e033dfb9d166440a135/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/impl/LayerImpl.java#L1566

Due to implementation of EDataTypeUniqueEList and [javadoc ](https://download.eclipse.org/modeling/emf/emf/javadoc/2.6.0/org/eclipse/emf/ecore/util/EDataTypeUniqueEList.html) givs a hint that it also has significant performance impact

PR #364, opened 2019-12-22
#363 fixes memory leak in featureChanges list

fgdrf:
see issue #363 for details

### DIFF
--- a/plugins/org.locationtech.udig.project.tests/src/org/locationtech/udig/project/internal/LayerImplTest.java
+++ b/plugins/org.locationtech.udig.project.tests/src/org/locationtech/udig/project/internal/LayerImplTest.java
@@ -37,6 +37,8 @@ import org.locationtech.udig.project.tests.support.MapTests;
 import org.locationtech.udig.ui.tests.support.UDIGTestUtil;
 
 import org.eclipse.core.runtime.NullProgressMonitor;
+import org.geotools.data.FeatureEvent;
+import org.geotools.data.FeatureEvent.Type;
 import org.geotools.data.FeatureSource;
 import org.geotools.data.FeatureStore;
 import org.geotools.data.Transaction;
@@ -46,6 +48,7 @@ import org.geotools.feature.FeatureCollections;
 import org.geotools.feature.simple.SimpleFeatureBuilder;
 import org.geotools.geometry.jts.ReferencedEnvelope;
 import org.geotools.referencing.CRS;
+import org.geotools.referencing.crs.DefaultGeographicCRS;
 import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


fgdrf:
@nprigour Would you please hava a look at this changeset too? Much appreciated

nprigour:
Hi @fgdrf,
Could you provide some insight when these add methods are expected to be called while working with the udig map display and/or tools because I couldn't figure out which map activities utilize them.

fgdrf:
Sure, let me try..

Whenever a Layer with a FeatureSource has been added to the map this listener is informed on FeatureEvents (such as Feature added/removed/modified). It caches last recent events to calculate bounding boxes and similiar things

In scenario where I add hundreds of thousands elements over time the Memory Footprint increases and will never be free objects while Garbage Collection is running (these were still in internal collection).

I cannot say why `add` is called and `add` with index not, but this causes the memory leak

Hope this helps

nprigour:
Hi @fgdrf ,

I regret to say that I was not able to reproduce and check this. I tried with both a shapefile and a mysql DB layer store by adding/removing/modifying features (from the map editor) but both `public void add(...)` methods seems never to be called. I debug using a breakpoint, adding a print out statement but still without success. Only the `public List<FeatureEvent> getFeatureChanges() ` method seems to be called every time a commit or rollback occurs on the map.
Do I miss something?

fgdrf:
Is your ListenerAddingInterceptor registered in plugin.xml of `org.locationtech.udig.project` bundle? This registers the Listener to the layer. Whenever a FeatureSource detects changes this listener is notified.

We are using MemoryFeatureStore with  a SimpleFeatureCollection we do update constantly with evens from a message-service. So maybe this was the puzzle I didn't mentioned.

HTH

nprigour:
Aha that was the missing puzzle part! I tried with a _MemoryFeatureStore_ and there it seems that the `add()` method is indeed called. However this is done only when pressing commit button (and not on every edit action)  where an _org.geotools.data.BatchFeatureEvent_  is emitted and added to the list of changes. By doing a step by step debug I figure out that the following sequence of actions occur on commit press:
1.  `getFeatureChanges().add(Object arg0)` of LayerImpl class is called as already mentioned
2.   `commitTransaction() `of EditManagerImpl is called which includes the following code (lines 380-386) where the featureChanges are cleared from every layer if present
```
        try {
            transaction.commitInternal();
            for (Layer layer : getMapInternal().getLayersInternal()) {
                if (layer.getFeatureChanges().size() != 0)
                    layer.getFeatureChanges().clear();
            }
        }
```
So essentially it seems that indeed on every commit the changes are added but eventually also cleared and the list does not ever unboundly increases. 
Of course here a couple of questions arose:
1. Why only one _org.geotools.data.BatchFeatureEvent_ is emitted during commit? 
2. Why `getFeatureChanges().add(Object arg0)` is not called for shapefiles and JDBC layers (both sources inherit from ContentFeatureSource)?

For the first question item the answer lies to the fact that geotools for transactions that are in non AUTO_COMMIT mode generate and populate a single _BatchFeatureEvent_ which is fired during commit.
For the second question item I do not have yet an answer. I figure out that the BatchFeatureEvent is indeed created but the `FeatureListener featureListener`  (see LayerImpl class at line 264) is not in the list of listeners to be informed during the commit action. Why this happens it really needs further  investigation.

fgdrf:
I remember we have a auto-commit DataStore setup and therefore on every add an Event is triggered. And the cause of the leak was this Listerner Implementation. So I'd like to proceed and merge it back to master if you don't have any objections. IMHO the Listener should have the same behavior for both add-methods and therefore its a straight forward fix (from code perspective) and might raise other question from user perspective we can discuss on the [developer mailing list](https://accounts.eclipse.org/mailing-list/udig-dev)

nprigour:
Why do you say that it is a straight forward fix? It sounds peculiar that during an addition we may also clear the list elements first! Normally I would expect to modify `add(int index, Object object)` to be align with `add(Object arg0)`. What if for example you added 12 events and while expecting to get them all at some point you obtain just 2?
After all the method is accesible only via the `getFeatureChanges()` and since _EDataTypeUniqueEList_  has a public `clear()` method you can call it at any time you want in application code. 
Based on current behavior of UDIG of course the proposed change seems not to have a real impact but wouldn't it be more rational to perform this size check somewhere else?  I.e within `public void changed(FeatureEvent featureEvent) {...}` block somewhere before or after the call to `getFeatureChanges().add(featureEvent)` at line 326 (if it really needs to be added)?
Moreover the defined size of 10 IMHO seems quite small to justify a clear action.

fgdrf:
Your are totally right, I don't say that the use case is correct nor that it makes any sense to clean up to a level of 10 (why not 100 or even 1000)? 
This fix IMHO adresses only the problem that on the one hand it cleans (who ever calls it) and on the other hand it doesn't (which leads to a memory leak in this specific case).

So it doesn't fix a possible wrong behavior ;) We can discuss this later on and hopefully we get an idea what intentions is behind. I guess we need to add at least more semantic e.g. and extract magic number "10" and give this variable a meaningful name (I don't have at the moment).

If we find out this is not needed anymore we can remove it ;)

nprigour:
From a quick inspection it seems that this add method is called just once at the line 326 of LayerImpl class that I mentioned above. 
Moreover whatever action you take for this PR we should generate another one related to how we can handle FeatureEvents based on the current behavior of the udig framework for every type f layer. As it is know it seems that we should change a bit the implementation of `ListenerAddingInterceptor` and the way Layer.featureListener member is created to include the `UDIGTransaction` of the EditManager as argument.

PR #367, opened 2019-12-25
Message bubble display access fix

fgdrf:
see origin pull request #300


nprigour:
Hi @fgdrf,

I cannot understand the relation of this one with #300. It does not seems to address the issue described there. Can you explain a bit?

fgdrf:
@nprigour Wondering a bit and referenced orgin commit ;)

PR #381, opened 2020-04-25
removes unsed xerces dependency

fgdrf:


fgdrf:
My read of https://github.com/geotools/geotools/commit/02bf28fe1f0a4678d99472771aeb6750175281db#diff-f21c80bd78a08efffe97d50946f77eaf is that xsd-core and therefore all GeoTools XML Parsers do not rely on Apache Xerces anymore

Excpet of one little village in netCDF land: gt-netcdf jar has a dependency to jaxen:jaxen:1.1.6 which has antransitive dependency to 
```
    <dependency>
      <groupId>xerces</groupId>
      <artifactId>xercesImpl</artifactId>
      <version>2.6.2</version>
      <scope>provided</scope>
    </dependency>
```

Which leads me to the question : Would we loose netCDF support if removing dependency to Bundle org.apache.xerces?

Can someone help testing with netCDF Dataesets?

fgdrf:
stale pull request, closing it. Whenever necessary I'll create another one

Issue #397, opened 2020-07-04
hasSelection & layerHasSelectionProperty - remove duplicate implementation


fgdrf:
class org.locationtech.udig.tool.select.internal.LayerHasSelectionProperty (Property hasSelection) duplicates functionality from org.locationtech.udig.project.ui.operations.LayerSelectionProperty (layerHasSelectionProperty)

In addition to this a namespace (e.g. org.locationtech.udig.ui.objectProperty as prefix) should be used to avoid conflicts with other properties

fgdrf:
relates to #391

PR #400, opened 2020-07-06
#397 unifies objectProperty for layer selection

fgdrf:


sschulz92 (changes requested):
Please convert the easymock test case into a Mockito test case!

Issue #398, opened 2020-07-04
Layer added to map is "inactive" in Layers View


fgdrf:
Whenever a new layer has been added to the map the layer is not displayed as selected in Layers View

From user perspective it is a bit confusing since the semantic is that content of a layer should not be displayed in Map. But content is shown.

activating the layer afterwards doesn't change anything

![image](https://user-images.githubusercontent.com/644229/86520629-0986b500-be47-11ea-93a6-72ad1c343433.png)

Expectation:
layer is active after added to map and layers view

PR #402, opened 2020-08-07
Issue #398 : valid state for layer in Layers View

fgdrf:
after a layer has been added (was inactive but visible)

Signed-off-by: Abderrahim Kassid <kassid@web.de>
Change-Id: I7fa3fcca6b4760cb34fc3f1dcc887a0bc9fb20cb


fgdrf:
tested it locally and it looks good. layers are added and are checked in layers view correctly! many thanks for this fix @kassid !

PR #414, opened 2020-09-03
moved style UI into separate bundle

fgdrf:
to allow to rollout an application without style perspective and style view (sdk-users) . 

Change style actions are still available (e.g. in Layers View), if providing bundles such as style.sld are deployed.

With this refactoring extension-processing has been refactored a bit to access attributes of Configuration Elements with constants definded in IStyleConfigurator, moved it to exported package org.locationtech.udig.style)

Moved also StyleLayer from internal package to exported package org.locationtech.udig.style

Change-Id: Ifdc2c50fcfd898ff45780094fa4947b364210d33


fgdrf:
@moovida can we go through this changeset together? Would be great to your point of view!

moovida (approved):
Hi @fgdrf , it looks to me that this is 99% simple refactoring and formatting, right? I saw also a few attributes extracted to more generic Style classes, which looks good to me. 

Is there something in particular you wanted me to look at?

fgdrf:
> 
> 
> Hi @fgdrf , it looks to me that this is 99% simple refactoring and formatting, right? I saw also a few attributes extracted to more generic Style classes, which looks good to me.
> 
> Is there something in particular you wanted me to look at?

Thanks for your review! You are right and I don't have anything else to add. Going to merge it into master soon :)

Issue #421, opened 2020-09-24
Allow to disable CRS Contribution Item to choose another Coordinate Reference System for the map


fgdrf:
In case uDig is intergrated in a RCP application, the CRS should be fix and not changable by the User it should be possible to disable the action to switch CRS for the map.

PR #422, opened 2020-09-24
Feature #421 : allows to disable CRS selection in Map

fgdrf:
with an additional option its possible to disable CRS Selection Dialog (CRS button in statusbar).

grafik

I haven't added an option in preferences dialog, since it should be a configuration to start uDig without having the capability to change CRS for the map.

BUT: What happens if the user opens an empty map and adds a resource to the map that provides a CRS the user doesn't want to have for the map? 

Change-Id: Idcffeb250f2dbb71d9ef1f29ddd3f98600b4294c

### DIFF
--- a/docs/build.xml
+++ b/docs/build.xml
@@ -89,9 +89,9 @@
     </target>
 
     <target name="sphinx" if="sphinx.available">
-        <echo message="Running sphinx-build -D release=${project.version} -b ${build} -d &quot;${build.directory}/${id}/doctrees&quot; . &quot;${build.directory}/${id}/${build}&quot;"/>
+        <echo message="Running sphinx-build -D release=${project.version} -W -b ${build} -d &quot;${build.directory}/${id}/doctrees&quot; . &quot;${build.directory}/${id}/${build}&quot;"/>
         <exec executable="sphinx-build" failonerror="true" dir="${basedir}/${id}">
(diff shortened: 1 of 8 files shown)


fgdrf:
> BUT: What happens if the user opens an empty map and adds a resource to the map that provides a CRS the user doesn't want to have for the map?

if defaultCRSPreference (config `org.locationtech.udig.project\defaultCRSPreference`) is not set, the layer that will be added first provides the CRS for the Map.

if defaultCRSPreference is set, e.g. `org.locationtech.udig.project\defaultCRSPreference=4326` then independently of layers CRS this default is used (in this case WGS84).

Users can change default CRS at runtime (Preferences Dialog -> Project -> Map) to preferred CRS. So if the button is disabled for CRS Selection, then its only possible to change CRS for new maps, not for existing maps. From UX perspective its not a good idea.

Ideas:
* if defaultCRSPreference is set and different to -1 then DISABLE_CRS_SELECTION will be used to decide
* if defaultCRSPreference  is -1 DISABLE_CRS_SELECTION is ignored and a warning is logged that describes how to configure combination of preferences correctly

Opinions?

fgdrf:
Going to submit this pull request soon, since udig behaves as before but has additional option to disable the button. We can adress the possible problem mentioned above later

PR #423, opened 2020-09-24
[HK] cleanup code regarding TilingRenderer 

fgdrf:
using one preferences constant for Tiled Rendering : P_TILED_RENDERING rather than two different.

deprecated P_TILING_RENDERING therefore ;)

Change-Id: I5188654f782a8aac31f67ec4260d4a31f2efb26a


fgdrf:
Hmm, just wondering that TilingRenderer is never used and TilingRendererTests is kind of empty.

I guess we can remove this experimental unused implementation, @jodygarnett @moovida would you agree?

PR #437, opened 2020-12-10
fix issiue 435

jaan-git:
File.renameTo method always return false, when the outFile is not exists; There is no need to check the lock when the output file does not exist.


fgdrf:
Thank you for providing a fix! 

Eclipse ECA check states the follwoing:


**Failures**

> Hu Jaan (2211844470@****.com) did not include the "Signed-off-by footer" which is required for all commits made by a contributor.

**Tips**

> Please check that each commit has the 'Signed-off-by' footer as part of the commit message. This can be added by checking out the offending commit and using git commit -a -s.
> 
> Additional information on contributing to Eclipse projects can be found on the Eclipse wiki under Development Resources/Handling Git Contributions, or under Development Resources/Contributing via Git for new committers.

I guess you can commit with --amend to rewrite commit. This requires to push with force to your branch ;)

PR #443, opened 2021-01-01
jfreechart libs as separate bundle

fgdrf:
to reduce complexity and dependency hell. It helps to have a single responsibility and might help for upgrade jfreechart easier.

For SDK-Users: They might use another version of jfreechart in their application or even without jfreechart based funtionality.


moovida (approved):
Looks good to me. Way cleaner :-)

PR #445, opened 2021-01-24
[HK] fix EMF Model (jts) and (re-)generated Code

fgdrf:
It changes namespace from com.vividsolution to org.locationtech in EMF Model.

Fixed Model regarding:
* runtime version
* emf runtime version (matches emf version from target-definition)
* several formatting settings 

### DIFF
--- a/plugins/org.locationtech.udig.project.edit/META-INF/MANIFEST.MF
+++ b/plugins/org.locationtech.udig.project.edit/META-INF/MANIFEST.MF
@@ -1,20 +1,24 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
-Bundle-SymbolicName: org.locationtech.udig.project.edit; singleton:=true
+Bundle-SymbolicName: org.locationtech.udig.project.edit;singleton:=true
 Bundle-Version: 2.3.0.qualifier
+Bundle-ClassPath: .
 Bundle-Activator: org.locationtech.udig.project.internal.provider.ProjectEditPlugin$Implementation
-Bundle-Vendor: udig.refractions.net
+Bundle-Vendor: %providerName
 Bundle-Localization: plugin
-Export-Package: org.locationtech.udig.project.internal.provider,org.lo
- cationtech.udig.project.internal.render.provider
+Export-Package: org.locationtech.udig.project.element.provider,
+ org.locationtech.udig.project.internal.provider,
+ org.locationtech.udig.project.internal.render.provider
 Require-Bundle: org.locationtech.udig.project;visibility:=reexport,
+ org.eclipse.emf.edit;visibility:=reexport,
+ org.eclipse.emf.ecore;visibility:=reexport,
  org.eclipse.jface,
  org.locationtech.udig.catalog,
  org.locationtech.udig.ui,
  org.eclipse.emf.edit.ui;visibility:=reexport,
  org.eclipse.core.runtime,
  org.locationtech.udig.catalog.ui,
- org.eclipse.emf.ecore.edit
+ org.eclipse.emf.ecore.edit;visibility:=reexport
 Bundle-ActivationPolicy: lazy
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
--- a/plugins/org.locationtech.udig.project.edit/plugin.properties
+++ b/plugins/org.locationtech.udig.project.edit/plugin.properties
@@ -1,6 +1,6 @@
 
 pluginName = Project Edit Support
-providerName = www.example.org
+providerName = udig.refractions.net
 # ====================================================================
 _UI_PropertyDescriptor_description = The {0} of the {1}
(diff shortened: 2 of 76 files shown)


erdalkaraca (approved):
most (or all) of the code is due to code generation and/or formatting... I only reviewed the ecore/genmodel files

fgdrf:
> most (or all) of the code is due to code generation and/or formatting... I only reviewed the ecore/genmodel files

@erdalkaraca Great, thanks for your review . I stumbled about MapImpl.getColorScheme() method where regenerated code looked like this:

`  /**
     *  
     * @generated
     */
    @Override
    public ColourScheme getColourScheme() {
        return colourScheme;
    }
`

and therefore I've added a **NOT** right behind `@generated` to avoid removing lazy initialization code.

erdalkaraca:
In that case I would also comment why you need lazy initialization. If I remember correctly, some code in uDig uses OS resources that should not be initialized immediately

> > most (or all) of the code is due to code generation and/or formatting... I only reviewed the ecore/genmodel files
> 
> @erdalkaraca Great, thanks for your review . I stumbled about MapImpl.getColorScheme() method where regenerated code looked like this:
> 
> ```
>    /**
>      *  
>      * @generated
>      */
>     @Override
>     public ColourScheme getColourScheme() {
>         return colourScheme;
>     }
> ```
> 
> and therefore I've added a **NOT** right behind `@generated` to avoid removing lazy initialization code. Now it looks as follows:
> 
> ```
>     /**
>      *  
>      * @generated NOT
>      */
>     @Override
>     public ColourScheme getColourScheme() {
>         if (colourScheme == null) {
>             colourScheme = ColourScheme.getDefault(getColorPalette());
>         }
>         return colourScheme;
>     }
> ```

fgdrf:
> MapImpl

I had some trouble initializing Maps in a multi-threaded environment and provided a patch in the past: https://github.com/locationtech/udig-platform/commit/c16617d2bb0e874aa4acd3f39288ead1ee71e327#diff-27f9c02051f99a7cdd33a9d8c0fcc5b38249aa00fdab87c5f2e630b411ee5cab

Question: Is it possible to define it in the model where and when initializing Content? Would it make sense to move the initialization to `org.locationtech.udig.project.internal.impl.ProjectFactoryImpl.createMap()` rather than in the MapImpl object itself?

erdalkaraca:
The factories just call the constructors, so there would be no benefit here. It is also not recommended to have code in the constructor of the generated classes. You can add the simple code in the ecore file by defining an ecore genmodel annotation on the attribute.

> > MapImpl
> 
> I had some trouble initializing Maps in a multi-threaded environment and provided a patch in the past: [c16617d#diff-27f9c02051f99a7cdd33a9d8c0fcc5b38249aa00fdab87c5f2e630b411ee5cab](https://github.com/locationtech/udig-platform/commit/c16617d2bb0e874aa4acd3f39288ead1ee71e327#diff-27f9c02051f99a7cdd33a9d8c0fcc5b38249aa00fdab87c5f2e630b411ee5cab)
> 
> Question: Is it possible to define it in the model where and when initializing Content? Would it make sense to move the initialization to `org.locationtech.udig.project.internal.impl.ProjectFactoryImpl.createMap()` rather than in the MapImpl object itself?

fgdrf:
> The factories just call the constructors, so there would be no benefit here. It is also not recommended to have code in the constructor of the generated classes. You can add the simple code in the ecore file by defining an ecore genmodel annotation on the attribute.

Cool, had a look at and tried it:

```
    <eOperations name="getColourSchema" eType="#//ColourScheme">
      <eAnnotations source="http://www.eclipse.org/emf/2002/GenModel">
        <details key="body" value="if (colourScheme == null) {&#xA;            colourScheme = ColourScheme.getDefault(getColorPalette());&#xA;}&#xA;return colourScheme;"/>
      </eAnnotations>
    </eOperations>
```

makes the trick and I could remove "NOT" from @generated. Question for me: Code is persisted in the model and might not fit PMD/Spotbugs rules in the future anymore. Developers need to know that is comming from the model, might reduce maintainability.

And las but not least, I guess its not a good practice to have some methodes bodies in the model and others implemented after generation of code and manually edited.

I guess for now I'll leave it as it is in the pull request but I learned how to do it the other way. Thanks for that!

> 
> > > MapImpl
> > 
> > 
> > I had some trouble initializing Maps in a multi-threaded environment and provided a patch in the past: [c16617d#diff-27f9c02051f99a7cdd33a9d8c0fcc5b38249aa00fdab87c5f2e630b411ee5cab](https://github.com/locationtech/udig-platform/commit/c16617d2bb0e874aa4acd3f39288ead1ee71e327#diff-27f9c02051f99a7cdd33a9d8c0fcc5b38249aa00fdab87c5f2e630b411ee5cab)
> > Question: Is it possible to define it in the model where and when initializing Content? Would it make sense to move the initialization to `org.locationtech.udig.project.internal.impl.ProjectFactoryImpl.createMap()` rather than in the MapImpl object itself?

PR #453, opened 2021-05-19
fix for metric rating computation

egouge:
Signed-off-by: Emily Gouge <egouge@refractions.net>

The getMetric() function states "A value of 1 = complete and 0 = incomplete, a value in-between represents a partial completeness."  The way this was originally being computed values > 1 were being returned if it was incomplete.  This fix flips the ratio and returns a value < 1 if the file doesn't have a CRS or Envelope.


fgdrf:
Great catch, fix looks good. Are you kind and would contribute a test-case as well?

egouge:
Frank - I'm trying really hard to add a test case and not getting very far.  Perhaps you can help -  I have a very simple class I added to the org.locationtech.udig.catalog.tests project:

`File file = new File("data/raster.tif");
Map<String, Serializable> map = new HashMap<String, Serializable>();
map.put("url", file.toURI().toURL());

IServiceFactory serviceFactory = CatalogPlugin.getDefault().getServiceFactory();
List<IService> services = serviceFactory.createService(map);
FakeProgress monitor = new FakeProgress();

for (IService service : services) {
   System.out.println(service.getInfo(monitor).getMetric());
}`

After adding the catalog.rasterings and catalog.worldimage to the set of required plugins the set of services being returned is empty.  It should return something.

egouge:
I figured out the issue above - "url" needs to be "URL".  I'll see if I can get the rest working.

egouge:
Try that - let me know if it doesn't work or is too complicated.

fgdrf (changes requested):
see pull request https://github.com/egouge/udig-platform/pull/8 for light-wight testing. See my comment in pull request

fgdrf (approved):

fgdrf:
@egouge Thanks for merging changes, going to squash it into master since build-feedback is ✅

fgdrf:
@egouge Thanks a ton for your contributions!

PR #460, opened 2021-08-28
[HK] experimental feedback for Java 11

fgdrf:


fgdrf:
@olsmn @kassid just have a look at additional feedback with Java 11 (still compile/target-level 1.8)

PR #462, opened 2021-08-31
Moves netcdf dependencies into (optional) fragment

fgdrf:
To reduced dependencies from libs.bundle that are very specific this is kind of a Proof of Concept 

Idea:

Core dependencies are managed with udig.libs bundle whereas specific "extensions" from geotools are fragments to udig.libs host-bundle. Cool thing in Eclipse OSGi : is only one classpath from callers perspectives :)

We did it before with jfreechart dependencies (see #443) and the benefit is : Whoever integrates uDig from udig.sdk can choose which extension to configure for final product. This leads to smaller footprints of uDig product for butter-and-bread solutions whereas all other extensions are additions lead to larger installations.

I belief this would also work for all other catalog configurations we have in the codebase, so see this pull as a Proof of Concept.

Due to the fact netcdf is loaded with JNA its not possible to add natives libs as fragments for specific OS platforms such as gtk.liinux.x86_64 or wind32.win32.x86_64. However native libs still need to be configured in PATH environment variable.

Feedback is welcome :)


plugins/org.locationtech.udig.libs.netcdf/pom-libs.xml
```diff
@@ -0,0 +1,231 @@
...
+			<version>1.1.3</version>
+			<scope>provided</scope>
+		</dependency>
+	
+		<!-- geotools library -->
+		<dependency>
+			<groupId>org.geotools</groupId>
+			<artifactId>gt-netcdf</artifactId>
+			<version>${geotools.version}</version>
+			<exclusions>
```
fgdrf, line 69: add a hint right here why these dependencies are excluded

plugins/org.locationtech.udig.libs.netcdf/pom-libs.xml
```diff
@@ -0,0 +1,231 @@
...
+
+	<artifactId>pom-libs.download.geotools.netcdf</artifactId>
+	<name>uDig NetCDF Core libraries Download</name>
+
+	<packaging>pom</packaging>
+
+	<properties>
+		<!-- see geotools root pom for details -->
+		<geotools.version>22.1</geotools.version>
+		<imageio-ext.version>1.3.2</imageio-ext.version>
```
fgdrf, line 29: move this version properties to root-pom-libs.xml to have single source to change in case of upgrading GeoTools or other dependencies

plugins/org.locationtech.udig.libs.netcdf/pom-libs.xml
```diff
@@ -0,0 +1,231 @@
...
+		<dependency>
+			<groupId>javax.media</groupId>
+			<artifactId>jai_imageio</artifactId>
+			<version>1.1</version>
+			<scope>provided</scope>
+		</dependency>
+		<dependency>
+			<groupId>javax.media</groupId>
+			<artifactId>jai_codec</artifactId>
+			<version>1.1.3</version>
```
fgdrf, line 54: use version property as of other dependencies (see also host-bundle)

fgdrf:
@olsmn @kassid Would be great if you can have a look at as well :)

features/org.locationtech.udig_sdk-feature/feature.xml
```diff
@@ -36,4 +36,13 @@ License v1.0 (http://www.eclipse.org/org/documents/edl-v10.html).
          id="org.locationtech.udig_language-feature"
          version="0.0.0"/>
 -->
+   <!-- adds optional netcdf libs bundle -->
+   <plugin
```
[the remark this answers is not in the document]
fgdrf: added fragment to unstable feature which is parte of update site (if build with profiles (`-Punstable,site`)

sschulz92:
As we had a look together 👍

PR #466, opened 2021-09-03
[HK] Fixes NPE in case any Exception

sschulz92:
while creating classes defined in extension

Signed-off-by: Sebastian Schulz <bastie92_spam@gmx.de>


plugins/org.locationtech.udig.ui.tests/src/org/locationtech/udig/internal/ui/operations/OpFilterPropertyValueConditionTest.java
```diff
@@ -0,0 +1,119 @@
+package org.locationtech.udig.internal.ui.operations;
```
fgdrf, line 1: since it is a new class file we should add a file comment as defined in https://github.com/locationtech/udig-platform/blob/master/extras/org.locationtech.udig.dev/codetemplates.xml here

fgdrf:
Commit with test-case looks good, going to merge it soon

fgdrf:
Thanks for your contribution! Looking forward for the other pull request in the queue :)

PR #469, opened 2021-09-06
[HK] DropAction checks MapPart instead of MapEditorPart

fgdrf:
to support ViewParts Implementations as well

Change-Id: Ic01ff5aa08c903d543eb62cf7509b0690e1a8c83


sschulz92:
Looks good to me 👍

PR #470, opened 2021-09-06
[HK] removes dependency to MapEditorPart in PlatformGeoSelectionManager

fgdrf:
rather than MapEditorPart to allow to use Listener for Views too


sschulz92:
This PR is doing more than just changing the interface, you may think about changing the PR title?

fgdrf (commented):
looks good

sschulz92 (approved):
Looks good to me 👍

PR #472, opened 2021-09-06
[HK] Trivial refactorings to module 'org.locationtech.udig.catalog.ui'

sschulz92:


plugins/org.locationtech.udig.catalog.ui/plugin.properties
```diff
@@ -42,7 +42,7 @@ description = Create a new in-memory service
 import.action.name = Catalog Import Drag and Drop
 
 layerExport.desc = Export all the features in a layer/resource.  The new layer is added to the catalog for later use.
-layerExport.name = Resource to Shapefile
+layerExport.name = All Features to Shapefile
```
fgdrf, line 45: If we rename this we should care about the user-docs as well (see markdowns in docs folder).

For example have a look at the the current generated docs (using Python and Sphinx):
* https://udig.github.io/docs/user/reference/Resource%20to%20Shapefile%20Wizard.html
* https://udig.github.io/docs/user/getting_started/walkthrough2/ExportingtoShapefile.html

plugins/org.locationtech.udig.catalog.ui/src/org/locationtech/udig/catalog/ui/export/CatalogExportWizard.java
```diff
@@ -294,6 +296,31 @@ public void run() {
         return true;
     }
 
+    private SimpleFeatureType removeUnsupportedTypes(final SimpleFeatureType destinationFeatureType) {
```
fgdrf, line 299: Aha, looks like removal of special (generic) attributes comming from feature-type. Attributes with type Object are ignored and a message is sent to System.out

I would appreciate it to log a warning message (wich would occur in errorlog-view. A comment for this private method would be great too :)

Is it possible to provide a testcase for such a constellation? What would happen if this methoud would not exists? Is the export broken?

fgdrf (changes requested):

plugins/org.locationtech.udig.catalog.ui/plugin.properties
sschulz92: I updated the documentation in another commit!

docs/user/en/reference/Resource to Shapefile Wizard.rst
```diff
@@ -1,4 +1,4 @@
-Resource to Shapefile Wizard
```
fgdrf, line 1: Wonderung if it would be neccessary to rename file either. Might leed to update references as well

docs/user/en/getting_started/walkthrough2/ExportingtoShapefile.rst
```diff
@@ -5,7 +5,7 @@ In addition to editing content from many sources, you can also export content to
 
 * With **Tasmania roads** selected, choose :menuselection:`File --> Export` from the menu bar
 
-* Select the **Resource to Shapefile** and press :guilabel:`Next`
+* Select the **All Features to Shapefile** and press :guilabel:`Next`
 
   .. image:: images/100000000000020D000002265A8298F4.png
```
fgdrf, line 10: I guess the image still shows "Resource to Shapefile", just open uDig and take a screenshot to replace image ;)

docs/user/en/reference/Resource to Shapefile Wizard.rst
fgdrf: see ouput of the build-log:

```
 [exec] /home/runner/work/udig-platform/udig-platform/docs/user/en/reference/Export Wizard.rst:12:toctree contains reference to nonexisting document 'reference/All Features to Shapefile Wizard'
 ```

fgdrf (changes requested):

docs/user/en/getting_started/walkthrough2/ExportingtoShapefile.rst
fgdrf: @sschulz92 going to replace the image soon and remove test-case for now

PR #473, opened 2021-09-06
[HK] Refactorings to tool modules

sschulz92:


sschulz92:
Will re-open the PR later again.

PR #474, opened 2021-09-07
remove Export-Package org.h2 from MANIFEST.MF

kassid:
Change-Id: If482a5ed24e44811b1b267f32e7e32b93911ce7d


fgdrf:
since it is not required in uDig itself this changeset if fine to avoid conflicts with other bundles from my point of view. 
```
[INFO] +- org.geotools.jdbc:gt-jdbc-h2:jar:22.1:compile
[INFO] |  +- org.opengeo:geodb:jar:0.9:compile
[INFO] |  |  \- net.sourceforge.hatbox:hatbox:jar:1.0.b10:compile
[INFO] |  \- com.h2database:h2:jar:1.1.119:compile
```

PR #475, opened 2021-09-07
CQ12676 : use Eclipse Orbit bundle apache.commons.codec

kassid:
which avoids bundling commons-codec dependencies again into udig bundles.

Change-Id: Ic7483438df15c06f66ef0159057a96549935abcf


plugins/org.locationtech.udig.libs/META-INF/MANIFEST.MF
```diff
@@ -90,6 +89,7 @@ Import-Package: oracle.core.lmx;resolution:=optional,
  javax.media.jai.tilecodec,
  javax.media.jai.util,
  javax.media.jai.widget,
+ org.apache.commons.codec;version="1.9.0",
```
fgdrf, line 92: I'd suggest o add other exported packages from commons-codec as well, otherwise it might fail to NoClassDefFoundErrors

```
org.apache.commons.codec;version="1.2.0"
org.apache.commons.codec.binary;version="1.2.0"
org.apache.commons.codec.digest;version="1.2.0",
org.apache.commons.codec.language;version="1.2.0"
org.apache.commons.codec.language.bm;version="1.2.0"
org.apache.commons.codec.net;version="1.2.0"
```

and I'd prefare the version that is the minimun required version which is:

from dependency-tree, if common-codec is not managed as provided in pom-libs:

```
[INFO] +- org.geotools:gt-wms:jar:22.1:compile
[INFO] |  \- commons-httpclient:commons-httpclient:jar:3.1:compile
[INFO] |     \- commons-codec:commons-codec:jar:1.2:compile
```

plugins/org.locationtech.udig.libs/META-INF/MANIFEST.MF
```diff
@@ -13,8 +13,7 @@ Require-Bundle: org.eclipse.core.runtime;visibility:=reexport,
  org.apache.xerces;visibility:=reexport,
  org.apache.log4j;visibility:=reexport,
  org.eclipse.xsd;visibility:=reexport,
- org.eclipse.swt;visibility:=reexport,
- org.apache.commons.codec;bundle-version="1.9.0";visibility:=reexport
```
fgdrf, line 17: good idea not tu use required bundle. others might have a never version of from another provider. This makes live easier from integration point of view :)

fgdrf:
see CQ #12676:  https://dev.eclipse.org/ipzilla/show_bug.cgi?id=12676

fgdrf:
Looks great! Thanks for contributing this improvement! It reduces installation footprint a bit. Looking forward for #479

PR #476, opened 2021-09-07
[HK] remove unused dependency to gt-css (udig.libs)

ybnkv:
Signed-off-by: Yvan Benekov <ybnkv@gmx.de>


sschulz92 (approved):
Looks good to me 👍

plugins/org.locationtech.udig.libs/.classpath
```diff
@@ -114,7 +110,6 @@
 	<classpathentry exported="true" kind="lib" path="lib/gt-xsd-wmts-22.1.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/gt-xsd-wps-22.1.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/h2-1.1.119.jar"/>
-	<classpathentry exported="true" kind="lib" path="lib/hatbox-1.0.b11.jar"/>
```
fgdrf, line 117: wondering why hatbox has been removed from classpath, which updated added recently with #458 

Please check, if the commit is based on the latest commit from master

plugins/org.locationtech.udig.libs/.classpath
```diff
@@ -150,6 +145,7 @@
 	<classpathentry exported="true" kind="lib" path="lib/jaxb-api-2.4.0-b180830.0359.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/jaxb-runtime-2.4.0-b180830.0438.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/jdom2-2.0.6.jar"/>
+	<classpathentry exported="true" kind="lib" path="lib/jfreechart-1.0.13.jar"/>
```
fgdrf, line 148: looks like this dependency never made it into classpath file but is required for oms:

from `mvn dependency:tree -f pom-libs.xml`

```
[INFO] +- org.jgrasstools:jgt-oms3:jar:0.7.6:compile
[INFO] |  \- jfree:jfreechart:jar:1.0.13:compile
```

however, if it is required, then it should be added to bundle-classpath in MF-File as well

fgdrf (changes requested):

fgdrf:
@ybnkv Currently I'm working on a fix for jfreechart dependencies (we moved with #443). two different dependencies seems to rely on jfreechart modules:

from udig.libs\pom-libs::
```
[INFO] +- org.geotools:gt-charts:jar:22.1:compile
[INFO] |  \- jfree:eastwood:jar:1.1.1-20090908:compile
[INFO] |     \- jfree:jcommon:jar:1.0.13:compile
```

and 
```
[INFO] +- org.jgrasstools:jgt-oms3:jar:0.7.6:compile
[INFO] |  \- jfree:jfreechart:jar:1.0.13:compile
```

so my plan: 
* set jfree:jcommon and jfree:jfreechart as provided dependency in libs-bundle
* import org.jfree packages in libs-bundle

PR #481, opened 2021-09-07
[HK] Remove JNA dependency

sschulz92:


fgdrf:
please have a look at libs.netcdf module which currently packages jna lib/jna-4.2.2.jar

to support java 11 we need to add a bundle, but the question is where to get 4.2.x from since Eclipse Orbit repository provides 4.1.0 "only" (see http://download.eclipse.org/tools/orbit/downloads/drops2/R20180330011457/repository)

Independently from that jgrasstools seems to required it:
```
[INFO] +- org.jgrasstools:jgt-oms3:jar:0.7.6:compile
[INFO] |  \- com.sun.jna:jna:jar:3.0.9:compile
```

@moovida can you describe the feature and workflow how to test jgrasstools in uDig with jna? Which libraries are required and how to configure uDig to run with these? Would it also run with newer version such as 4.1 or even 4.2?

Much appreciate your collaborration on this!

fgdrf:
Just an update, we already using a newer orbit repository (see https://github.com/locationtech/udig-platform/blob/master/extras/org.locationtech.udig.target/org.locationtech.udig.target.target#L115) which provides never versions of jna (4.5.1 and 5.6.0). I'm not sure how it works because for each version is a com.sun.jna.platform artifact with the same version available (Maybe API and Runtime implementation).

fgdrf:
@sschulz92 Created a pull-request for your branch which adresses to load jna from Orbit OSGi bundle rather than packaging maven dependencies to jna into libs bundles : https://github.com/sschulz92/udig-platform/pull/1

fgdrf:
see Eclipse CQ# 15797 : https://dev.eclipse.org/ipzilla/show_bug.cgi?id=15797

moovida:
Hi @fgdrf , I think JNA is used only for the network tools (i.e. attaching to epanet for water networks, which is in C). These are simple bindings, so I can't imagine there would be problems with upgrades.

PR #483, opened 2021-09-09
[HK] use IActionBars rather than specific IActionBars2

fgdrf:
Change-Id: Id440a85a3af3674fec752b1db47245ebb0996cb0


fgdrf:
@kassid any objections to merge it?

PR #486, opened 2021-09-13
Fix name of file and related files according to renaming

sschulz92:
…  of the classname


fgdrf:
Thanks for your contribution!

PR #492, opened 2021-09-17
Remove 'no javadoc' comments and fix several indention mixtures

sschulz92:
Around 50 files are missing, I will do this in a second PR though as this PR is already to big :( I am sorry.

Within this PR I automatically removed trailing whitespaces and updated the file-comment to follow the newest change in udig-platform master!


sschulz92:
I finished all files with my 2nd commit :)

fgdrf:
I'm wondering if we can do it on a module-level using clean-up tasks from within eclipse (`Window -> Preferences -> Java -> Code Style -> Clean up`) with [code-cleanup.txt](https://github.com/locationtech/udig-platform/files/7185917/code-cleanup.txt) saved as xml file and importet right there.
Note it doesn't fix the javadoc sections bot fixes things as follows:

grafik

which would improve code-base a lot and allows to refactor things with better support.

Opinions?

sschulz92:
Sounds decent, on the other hand there will be a lot of Pull-Request-Reviews incoming for you :D

PR #496, opened 2021-09-21
Avoid setting IntroManager standBy

sschulz92:
Change-Id: Ia4550b1210901f98fb0c2f8ad825b23d2d473625

Signed-off-by: Sebastian Schulz <bastie92_spam@gmx.de>


plugins/org.locationtech.udig.ui/src/org/locationtech/udig/internal/ui/UDIGStartup.java
```diff
@@ -22,62 +25,41 @@
...
+    public void earlyStartup() {
         checkForGDI();
         loadCommonlyUsedObject();
-		PlatformUI.getWorkbench().addWorkbenchListener(ShutdownTaskList.instance());		
-		standByWelcome(); // uDig wants the welcome page put to the side on startup
-	}
-	/**
-	 * Make sure the Welcome screen is put into standby mode; prevents
-	 * confusion when Add Layer is used and the map is displayed behind the welcome
-	 * page (also looks better).
```
fgdrf, line 67: Wondering if there is a way to decide if a layer has been added. Wait .. please have a look at IStartup Implementation of project.ui plugin (StartupOpenMaps).

_If the DND extension point will be called so that it will act as if the url/string whatever was
dropped on the application. So if a url to shapefile is on the commandline a new map will be
openned and will display the shapefile._

I guess we should consider to move implementation code from udig.ui to project.ui. Or provide the functionality still from udig.ui plugin but used from project.ui plugin. This may also be configurable, e.g. a preference constant MINIMIZE_WELCOME_ON_DATA_URL or something like that.

fgdrf: I had a look at UDIGApplication which is the only class which checks obviously for programm-Arguments (`Platform.getCommandLineArgs()`) . I expected there an option to add layer on startup but couldn't find a hint.

However, socket-plugin provides functionality to add layer resources using message-header `DisplayLayer version:1.0`.

I believe its only option ath the time of writing, I'd suggest remove `standByWelcome()`as provided in this pull request. Also because of it changes default behavior in eclipse RCP applications where the Welcome-Screen is essential to guide the first-timw-user through the application.

Issue #497, opened 2021-09-22
Switch to compile-level 11


sschulz92:
As Java is moving forward we might think about upgrading the uDig platform to use a newer compile level. If we decide to move on to Java 11 we may also think about increasing the uDig version to 3.0.0 :)

@fgdrf what do you think about this issue? Is this feasible to work on?

PR #502, opened 2021-09-22
upgrade Eclipse Tycho to 2.3 as preperation to move on to java11

fgdrf:
as request in issue #497 this pull-request adresses an upgrade to Tycho 2.3.0

### DIFF
--- a/.github/workflows/maven-java-linux.yml
+++ b/.github/workflows/maven-java-linux.yml
@@ -16,10 +16,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up JDK 8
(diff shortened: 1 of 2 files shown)


fgdrf:
Looks like it is the other way aroung, tycho 2.3 requires at least java 11 as jdk for the build ..

sschulz92:
I am wondering if we want to do such a change in uDig 2.3.0 or to release the existing repository to a new major version which requires Java 11 or newer.

fgdrf:
You are right, lets pospone for 2.3.0 We can re-open the pull for 3.0.0 and upgrade to Java 11

PR #503, opened 2021-09-22
Findbugs to spotbugs

fgdrf:
allow to check code with spotbugs rather than retired findbugs plugin.

just run build with `spotbugs:check` to get high-ranked issues in the codebase. At the moment of writing ist configured not to fail, just report high level problems.


sschulz92 (approved):
👍

PR #511, opened 2021-09-24
Add 'Automatic-Module-Name' to all MANIFEST.MF

sschulz92:
This new attribute is required to move on to Java 9 or newer.


fgdrf:
Not sure about this one at the moment since compile-level and source-level is still Java 1.8.

And I remember some clitches Michael mentioned I'd like to clarify first ;)

PR #519, opened 2021-09-30
Remove deprecated method 'getGeoResource' from 'ILayer'

sschulz92:


plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/ILayer.java
```diff
@@ -72,59 +70,52 @@
...
-     * </code></pre>
-     * 
+     * </code>
+     * </pre>
+     *
      * </p>
-     * 
+     *
      * @param clazz class of the resource that the IGeoResource claims it can adapt to.
      * @return true if a IGeoResource exists that canResolve to resourceType.
```
fgdrf, line 108: the javadoc for retrurn is obviously wrong, should be 'the first found geoResource that can resolve to clazz.'
Are you kind to adress this as well?

sschulz92: I just pushed a second commit, is this issue resolved than?

PR #520, opened 2021-09-30
Remove deprecated class 'Identifier'

sschulz92:


plugins/org.locationtech.udig.catalog.ui/src/org/locationtech/udig/catalog/ui/CatalogUIPlugin.java
```diff
@@ -314,51 +314,52 @@ public ImageData getImageData() {
...
-        if( isFeature ){
-        	return ISharedImages.FEATURE_OBJ;
-        }
-        else {
-        	return ISharedImages.GRID_OBJ;
+        if (isFeature) {
+            return ISharedImages.FEATURE_OBJ;
+        } else {
+            return ISharedImages.GRID_OBJ;
         }
```
fgdrf, line 363: I need to check if implementation is the same, since Identifier has not delegated to ID implementations. And I'm going to test it in the final product as well, may take awhile

fgdrf: what about check for isMemory() or isFeature only and default ist ISharedImages.GRID_OBJ:
```
private static String iconInternalResource(ID id, boolean isFeature) {
        if (isFeature) {
            return ISharedImages.FEATURE_OBJ;
        }
        if (id.isDecorator()) {
            return ISharedImages.GRAPHIC_OBJ;
        }
        if (id.isMemory()) {
            return ISharedImages.MEMORY_OBJ;
        }
        return ISharedImages.GRID_OBJ;
}
```

btw. implementation of isDecorator is slighty different to isGraphic, I miss the following:

```
        if ("mapgraphic".equals(PROTOCOL)){
            return true; // we are hitting some mangled url
        }
``` in ID implementation.

Comment (sschulz92) [reply plugins/org.locationtech.udig.catalog.ui/src/org/locationtech/udig/catalog/ui/CatalogUIPlugin.java:363]: In regards of `isGraphic()` I had a look into ID.java and I found the part you mentioned. Is this not correct?
```
/**
 * @return true if ID represents a decorator
 */
public boolean isDecorator() {
    if (url == null)
        return false;

    String HOST = url.getHost();
    String PROTOCOL = url.getProtocol();
    String PATH = url.getPath();
    if (!"http".equals(PROTOCOL)) //$NON-NLS-1$
        return false;
    if (!"localhost".equals(HOST)) //$NON-NLS-1$
        return false;

    if (!"/mapgraphic".equals(PATH)) //$NON-NLS-1$
        return false;
    return true;
}
```

Comment (fgdrf) [reply plugins/org.locationtech.udig.catalog.ui/src/org/locationtech/udig/catalog/ui/CatalogUIPlugin.java:363]: In Identifier isGraphic looked like this:
```
   /**
     * @deprecated Please use new ID( url ).isGraphc()
     * @param url
     * @return true if url is a graphic url
     */
    public static boolean isGraphic( URL url ) {
        // http://localhost/mapgraphic
        String HOST = url.getHost();
        String PROTOCOL = url.getProtocol();
        String PATH = url.getPath();
        if ("mapgraphic".equals(PROTOCOL)){
            return true; // we are hitting some mangled url
        }
        if (!"http".equals(PROTOCOL))return false; //$NON-NLS-1$
        if (!"localhost".equals(HOST))return false; //$NON-NLS-1$

        if (!"/mapgraphic".equals(PATH))return false; //$NON-NLS-1$

        return true;
    }
```

which checks right in the middle if **PROTOCOL** is mapgraphic whereas the ID implementation of isDecorator doesn't. I'd suggest to add

```
if ("mapgraphic".equals(PROTOCOL)){
    return true; // we are hitting some mangled url
}
``` 
in ID.isDecorator().

fgdrf (approved):

Issue #517, opened 2021-09-29
OSM Spatial Toolbox with JGrassTools 0.7.3 are incompatible


fgdrf:
loading `jgt-hortonmachine-0.7.3.jar` and `jgt-jgrassgears-0.7.3.jar` with spatial toolbox (see https://udig.github.io/docs/user/getting_started/Introduction%20to%20the%20Spatial%20Toolbox.html) doesn't work anymore. Its due to incompatibility of jts used in udig (com.locationtech.jts) to the required older version (com.vibidsolution.jts)

```
java.lang.NoClassDefFoundError: com/vividsolutions/jts/geom/Geometry
	at java.lang.Class.getDeclaredMethods0(Native Method)
	at java.lang.Class.privateGetDeclaredMethods(Class.java:2701)
	at java.lang.Class.privateGetPublicMethods(Class.java:2902)
	at java.lang.Class.getMethods(Class.java:1615)
	at oms3.util.Components.getComponentClasses(Components.java:360)
	at org.locationtech.udig.omsbox.core.OmsModulesManager.scanForModules(OmsModulesManager.java:281)
	at org.locationtech.udig.omsbox.core.OmsModulesManager.browseModules(OmsModulesManager.java:169)
	at org.locationtech.udig.omsbox.view.OmsBoxView$8.run(OmsBoxView.java:365)
	at org.locationtech.udig.ui.PlatformGIS$1$2$1.run(PlatformGIS.java:373)
	at org.locationtech.udig.ui.PlatformGIS.lambda$2(PlatformGIS.java:105)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.ClassNotFoundException: com.vividsolutions.jts.geom.Geometry
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 14 more
```

@moovida is there a newer version available which migrated to the newer org.locationtech.jts packages? Otherwise it looks like we should retire osmbox-module in uDig

moovida:
[...]

PR #532, opened 2021-10-02
#517 retires osmbox & jconsole (GeoScript) module

fgdrf:
due to incompatibility of jts versions


sschulz92:
Why did you move all sources to `archive` instead of complete deletion? Since we are using Git, you can easily revert the deletion in case you decide to use it later on again. I doubt that this will happen to be honest. In case you need something similar in the future, you will write a new feature with state-of-the-art tools at this time.

fgdrf:
> Why did you move all sources to `archive` instead of complete deletion? Since we are using Git, you can easily revert the deletion in case you decide to use it later on again. I doubt that this will happen to be honest. In case you need something similar in the future, you will write a new feature with state-of-the-art tools at this time.

If these are still visible it might be easier to involve others. I guess if they are deleted these are out of mind. And IMHO the jconsole is a great feature to allow scripting with datasources and map but right now the effort was to big from me to untie it from omsbox. However I personally believe that [GeoScript](https://geoscript.net) intergration should be enabled again in the near future.

That said, @moovida what do you think about collaborating on this topic?

moovida:
Hi @fgdrf , as much as I wish to, I do not have any project that involves uDig at the current time. With HM and SMASH I have 0 free time left to help on this. I am really sorry. I also believe a GIS needs to have spatial analysis features and scripting, but that doesn't help as a thought alone.

Also, the last times I downloaded the current uDig version, it wasn't usable on my mac.

fgdrf:
> Hi @fgdrf , as much as I wish to, I do not have any project that involves uDig at the current time. With HM and SMASH I have 0 free time left to help on this. I am really sorry. I also believe a GIS needs to have spatial analysis features and scripting, but that doesn't help as a thought alone.
@moovida don't mind, I totally understand you situation ass I heared on #FOSS4G your are involved in more than mentioned projects.

> Also, the last times I downloaded the current uDig version, it wasn't usable on my mac.

Not that good, however are you kind to report any issue you stumbling about. Tought one here that I do not have a Mac anymore to verify a fix is working on that platform.

PR #533, opened 2021-10-04
Remove unnessecary SuppressWarnings of type 'NLS'

sschulz92:


fgdrf:
Not sure if this is still valid since omsbox bundle moved to archive recently, would you please re-base and ignore omsbox changes?

sschulz92:
I removed the change, the other 5 files are still in production :)

PR #536, opened 2021-10-04
Replace deprecated ViewerSorter by ViewerComparator

sschulz92:


fgdrf:
Approved ✔️ 

I assume its a straight forward fix since all calls were delegates to Comparators and implementions in viewers are the same.

Thanks a ton cleaning deprecation warnings from Eclipse framework

PR #501, opened 2021-09-22
[DOC] addes Eclipse IDE getting started guide

fgdrf:
covers:
* basic installation steps
* preferences
* project setup and product launch

Signed-off-by: Frank Gasdorf <fgdrf@users.sourceforge.net>

PR #540, opened 2021-10-05
Apply code formatter to classes of module 'udig.tools.feature.util'

sschulz92:


fgdrf:
@sschulz92 thanks for providing this pull-request. I totally understand thats easier to maintain the code if you don't have formating clitches which causes merge issues and even more bad-readable code.

I'd prefer if we leave the code how it is, and change its only if required, e.g. for new features or bugfix-changes. Would you agree.

IMHO, pros for changing only on occasion:
* keep origin contributor as long as possible 
* "[boy-scout rule : Always leave the camp cleaner than you found](https://matheus.ro/2017/12/11/clean-code-boy-scout-rule/)"

Downside to keep the code as it is:
* codebase is kind of inconsitent regarding coding-rules (for a longer time) and might confuse others to getting started

Contra for boy-scout rule: its not trivial to see the real change because formatings just hide the fix. IMHO it makes it harder to merge (and backport bugfixes to older releases - which is not really a use-case for uDig :P)

Therefore we updated the Getting started guide recently with #501 and I'd suggest to change classes/packages/modules only on occassion.

Opinions?

PR #542, opened 2021-10-05
Re-add opencsv (v 3.7) to fix Eclipse errors

sschulz92:
After the merge of https://github.com/locationtech/udig-platform/pull/532 I faced issues accross the project because "opencsv" could not be found anymore. Some of the projects rely on this library though.

Therefore I added the library again and all issues are resolved within my Eclipse!


fgdrf:
Have you executed `mvn clean package -f pom-libs.xml` on root-checkouit directory? If executed in plugins/org.locationtech.udig.libs only, other 3rd-party libraries are not downloaded. And not to forget: Refresh modules in Package Explorer within Eclipse IDE afterwards ;)

PR #544, opened 2021-10-06
Apply code-formatter in module 'org.locationtech.udig.catalog'

sschulz92:
I fixed also the deprecated use of SubProgressMonitor in this module :)

### DIFF
--- a/plugins/org.locationtech.udig.catalog/src/org/locationtech/udig/catalog/internal/ServiceFactoryImpl.java
+++ b/plugins/org.locationtech.udig.catalog/src/org/locationtech/udig/catalog/internal/ServiceFactoryImpl.java
@@ -1,7 +1,7 @@
-/*
- *    uDig - User Friendly Desktop Internet GIS client
- *    http://udig.refractions.net
- *    (C) 2004-2012, Refractions Research Inc.
+/**
+ * uDig - User Friendly Desktop Internet GIS client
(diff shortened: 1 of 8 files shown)


plugins/org.locationtech.udig.catalog/src/org/locationtech/udig/catalog/internal/ServiceFactoryImpl.java
```diff
@@ -19,114 +19,112 @@
...
         return createService(target);
     }
 
     /** @deprecated use createService */
-    public List<IService> acquire( final URL id, final Map<String, Serializable> params ) {
+    @Deprecated
+    @Override
+    public List<IService> acquire(final URL id, final Map<String, Serializable> params) {
         return createService(params);
     }
```
fgdrf, line 87: Is there another pull request that removes deprecated aquire methodes from IServiceFactory?

fgdrf (approved):

plugins/org.locationtech.udig.catalog/src/org/locationtech/udig/catalog/internal/ServiceFactoryImpl.java
sschulz92: I have a branch which is doing this action but I stopped the PR as I need to sort things out!

PR #546, opened 2021-10-07
Delete class 'ProgressMonitorTaskNamer' and use SubMonitor instead

sschulz92:
The deleted class `ProgressMonitorTaskNamer` implemented the deprecated `SubProgressMonitor`. As far as I can see the class `SubMonitor` is doing exactly what the deleted class did, therefore the class is not required anymore we can just easily use the SubMonitor instead - this is what I did.


fgdrf:
because it made master build "red" and all other pull requste that got this changset merged. I had a quick look at the mentioned lines where tests failing and could not find out what might cause the problem.

PR #548, opened 2021-10-08
split build steps insto compile (parallel) & test

fgdrf:
while tests are executed non-parallel. Looks like xvfb has problems while maven forks processes


fgdrf:
Here we go, tests are fine.

Compile for all modules incl. tests and tutorials within ~3min.
Test execution ~9-10min

:)

PR #549, opened 2021-10-08
Apply code-formatter to Java classes of 'docs'

sschulz92:


fgdrf:
These are just support-tools to verify if docs are valid and toc for eclipse user-helps references all pages.

However, I guess it makes sence to have the same style like defined in formatter. @jodygarnett would you agree to merge it back since you're the initial contributor of these classes?

fgdrf:
I belive @jodygarnett is fine with this change, lets proceed :)

PR #553, opened 2021-10-08
adds Gitter chat channel

fgdrf:
.. for QA-Sessions and open discussions


fgdrf:
Gitter.im (Log-In with github account) we can use for open discussions for uDig or extented to LocationTech projects

PR #554, opened 2021-10-11
refactor udig catalog oracle

fgdrf:
to test spotbugs action, if pull-request is not comming from forked repositiory (see https://github.com/jwgmeligmeyling/spotbugs-github-action/issues/30)


plugins/org.locationtech.udig.catalog.oracle/src/org/locationtech/udig/catalog/internal/oracle/OracleGeoResource.java
```diff
@@ -161,63 +170,34 @@ protected OracleResourceInfo createInfo( IProgressMonitor monitor ) throws IOExc
...
+                        .log(new org.eclipse.core.runtime.Status(IStatus.WARNING,
+                                "org.locationtech.udig.catalog", 0, //$NON-NLS-1$
+                                Messages.OracleGeoResource_error_layerBounds, e));
                 bounds = new ReferencedEnvelope(new Envelope(), null);
             }
 
             icon = Glyph.icon(ft);
-            keywords = new String[]{"postgis", //$NON-NLS-1$
-                    ft.getName().getLocalPart(), ft.getName().getNamespaceURI()};
+            keywords = new String[] { "postgis", //$NON-NLS-1$
```
fgdrf, line 186: Hui, whats this? the class is within catalog.oracle module and keyword is postgis? I guess this is not valid .. and I'll create an issue for this

plugins/org.locationtech.udig.catalog.oracle/src/org/locationtech/udig/catalog/internal/oracle/OraclePlugin.java
```diff
@@ -23,115 +23,130 @@
...
+
+    /**
+     * This method is called when the plug-in is stopped
+     */
+    @Override
+    public void stop(BundleContext context) throws Exception {
+        plugin = null;
+        resourceBundle = null;
+        super.stop(context);
+    }
```
fgdrf, line 57: Hmm, wondering a bit since Eclipse generates Activator-class with:
```
	// The shared instance
	private static Activator plugin;
	
	/**
	 * The constructor
	 */
	public Activator() {
	}

	@Override
	public void start(BundleContext context) throws Exception {
		super.start(context);
		plugin = this;
	}

	@Override
	public void stop(BundleContext context) throws Exception {
		plugin = null;
		super.stop(context);
	}

	/**
	 * Returns the shared instance
	 *
	 * @return the shared instance
	 */
	public static Activator getDefault() {
		return plugin;
	}
```

I've checked such a class locally with Eclipse Spotbugs Integration but without any complains. @sschulz92 What kind of error/warning have you fixed with this change?

PR #559, opened 2021-10-11
Fix several possible NPE in 'org.locationtech.udig.image.georeferencing'

sschulz92:
Main change in inside file `CoordinateTableComposite.java` and is about possible NullPointerExceptions:
Method `getMainComposite()` can return NULL which is not verified internally. Therefore SonarQube complained that we need to verify if MainComposite is NULL or not.

SonarQube:
image


plugins/org.locationtech.udig.image.georeferencing/src/org/locationtech/udig/image/georeferencing/internal/ui/coordinatepanel/CoordinateTableComposite.java
```diff
@@ -79,1132 +69,1171 @@
...
-		};
+            @Override
+            public void capturedCoordinate(Coordinate newCoord) {
 
-		this.deletedListener = new DeletedCoordinateListener() {
+                addCoordinateFromMapToTable(newCoord);
 
-			public void deletedCoordinate(java.awt.Point point, InputEvent event) {
+                MainComposite mainComposite = getMainComposite();
+                assert mainComposite != null;
```
fgdrf, line 187: I guess this is the fix for potential NPE, quite hard to find due to re-format of code

sschulz92: There are many many rows like this which are changed. Hard to see while formatting, indeed.

PR #560, opened 2021-10-12
[HK] enables validation of libs-bundle setup, fixes issues

fgdrf:
adds two test-cases to check whether org.locationtech.uidg.libs bundle is configured correctly regarding:
* configured Bundle-Classpath entries where no such jar-file is bundled in lib-folder
* jars that are bundles but not configured in Bundle-Classpath
* black-listed jars (bundled elsewhere) are reported as well, e.g. jfreechart, omsbox, netcdf

with this changeset the problems are adressed as well found with test-cases

### DIFF
--- a/plugins/org.locationtech.udig.libs/pom-libs.xml
+++ b/plugins/org.locationtech.udig.libs/pom-libs.xml
@@ -73,20 +73,6 @@
 	</dependencyManagement>
 
 	<dependencies>
-		<!-- see org.locationtech.udig.libs.jfreechart -->
-		<dependency>
-			<groupId>jfree</groupId>
-			<artifactId>jfreechart</artifactId>
-			<version>1.0.13</version>
-			<scope>provided</scope>
-		</dependency>
-		<dependency>
-			<groupId>jfree</groupId>
-			<artifactId>jcommon</artifactId>
-			<version>1.0.13</version>
-			<scope>provided</scope>
-		</dependency>
-
 		<!-- uDig Specific Extras -->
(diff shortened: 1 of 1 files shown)


plugins/org.locationtech.udig.libs.tests/src/org/locationtech/udig/libs/tests/ManifestValidationTest.java
```diff
@@ -0,0 +1,123 @@
+package org.locationtech.udig.libs.tests;
```
sschulz92, line 1: What about the template of uDig?

plugins/org.locationtech.udig.libs/pom-libs.xml
```diff
@@ -74,6 +73,20 @@
 	</dependencyManagement>
 
 	<dependencies>
+		<!-- see org.locationtech.udig.libs.jfreechart -->
+		<dependency>
+			<groupId>jfree</groupId>
+			<artifactId>jfreechart</artifactId>
+			<version>1.0.13</version>
```
sschulz92, line 80: You could have used the property instead of the duplicated version here, right? But anyway, can stay as is from my perspective :)

sschulz92 (approved):
👍

plugins/org.locationtech.udig.libs/pom-libs.xml
fgdrf: created an issue for eastwood #562

PR #565, opened 2021-10-14
Remove class progress monitor task namer

sschulz92:
Second try ... see also the first PR: https://github.com/locationtech/udig-platform/pull/546

Locally my build is running fine - with tests! What can we do @fgdrf ?


fgdrf:
agree, tested locally (w/o parallel build option) and everything is fine. lets tes again mering it onto master

PR #573, opened 2021-10-21
[CQ-18796] using org.apache.commons.io from Orbit

fgdrf:
and exclude unused commons-math from libs-bundle

Change-Id: Ic39792cd55b6e1cd734a4efba21fa579df4a3e12


plugins/org.locationtech.udig.libs/.classpath
```diff
@@ -199,8 +197,8 @@
 	<classpathentry exported="true" kind="lib" path="lib/wkb4j-1.0-RC1.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/xml-commons-resolver-1.2.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/xpp3_min-1.1.4c.jar"/>
-	<classpathentry kind="lib" path="lib/commons-compiler-3.0.8.jar"/>
-	<classpathentry kind="lib" path="lib/gdal-2.2.0.jar"/>
-	<classpathentry kind="lib" path="lib/janino-3.0.8.jar"/>
+	<classpathentry exported="true" kind="lib" path="lib/commons-compiler-3.0.8.jar"/>
+	<classpathentry exported="true" kind="lib" path="lib/gdal-2.2.0.jar"/>
+	<classpathentry exported="true" kind="lib" path="lib/janino-3.0.8.jar"/>
```
fgdrf, line 202: @egouge maybe this missing `exported="true" option is the problem on master branch, why catalog.imageio cannot resolve gdal correctly. Can you check and verify this independently from this pull-request on master?

PR #581, opened 2021-11-01
using MapPart instead of MapEditor

fgdrf:
in preperation to remove mapEditor and as a transition from concrete implementation to interfaces this changset removes a left-over in a test-bundle.

Change-Id: I020c68b3d130fec189d9e07e024a5da7bc4932b7

### DIFF
--- a/plugins/org.locationtech.udig.info.tests/src/org/locationtech/udig/info/tests/TestViewportPane.java
+++ b/plugins/org.locationtech.udig.info.tests/src/org/locationtech/udig/info/tests/TestViewportPane.java
@@ -6,157 +6,183 @@ package org.locationtech.udig.info.tests;
 import java.awt.Dimension;
 import java.awt.image.BufferedImage;
 
+import org.eclipse.swt.graphics.Cursor;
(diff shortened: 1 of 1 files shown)


plugins/org.locationtech.udig.info.tests/src/org/locationtech/udig/info/tests/TestViewportPane.java
```diff
@@ -35,112 +34,139 @@ public TestViewportPane(Dimension displaySize) {
 		this.displaySize = displaySize;
 	}
 
-	public void addDrawCommand(IDrawCommand command) {
+	@Override
```
sschulz92, line 37: Indention seems to be wrong?

sschulz92 (changes requested):

plugins/org.locationtech.udig.info.tests/src/org/locationtech/udig/info/tests/TestViewportPane.java
fgdrf: fixed mix-up of tabs and spaces ;)

PR #586, opened 2021-11-02
Replace SubProgressMonitor with SubMonitor, remove unused class

sschulz92:
- Use `SubMonitor` instead of `SubProgressMonitor` in module **org.locationtech.udig.catalog.ui**
- Delete unused class `CountDownProgressMonitor.java`


fgdrf:
Looks good to me! Thanks a again for removing deprecations and move on to current Eclipse API and Implementions

PR #587, opened 2021-11-03
Extract function to format a Coordinate

sschulz92:


plugins/org.locationtech.udig.tool.default/src/org/locationtech/udig/tools/internal/CursorPosition.java
```diff
@@ -146,10 +145,10 @@ public void fill(Composite parent) {
             data.heightHint = 15;
             separator.setLayoutData(data);
             textArea = new Text(parent, SWT.BORDER | SWT.CENTER);
-            textArea.addKeyListener(this);
-            textArea.addFocusListener(this);
```
fgdrf, line 150: why are you removing key-listener & focus-listener? This allows the user to type a coordinate to center the viewport with SetViewportCenterCommand.

Otherwise the field is editable but without any effect on release/focus-lost.

PR #463, opened 2021-09-02
Fix several typos and trivial improvements across the project

sschulz92:

PR #588, opened 2021-11-03
Allows to abort edit tools

sschulz92:
with ESC & modifies user message shown on invalid geometry

Reproduction:

- Draw an invalid polygon on Map
- Error dialog appears
- You are not able to use ESC to exit the drawing


sschulz92:
Another change is the removal of IStatus in the error dialog. By removing it, the message is not shown twice anymore:

image

With my PR the message is shown only once.

fgdrf:
see releated pull request #463 where the "Ctrl-Z" has been removed from the message.

fgdrf:
Thanks for this fix, it really helps to abort the workflow.

PR #589, opened 2021-11-04
Allow to delete filter in layers

AndreB87:
It was not possible to delete a filter from a layer by removing it from the text field in the style editor. 

This pull request fixes the issue.

### DIFF
--- a/plugins/org.locationtech.udig.style.advanced.feature/src/org/locationtech/udig/style/advanced/lines/LinePropertiesComposite.java
+++ b/plugins/org.locationtech.udig.style.advanced.feature/src/org/locationtech/udig/style/advanced/lines/LinePropertiesComposite.java
@@ -52,13 +52,19 @@ public class LinePropertiesComposite implements ModifyListener, IStyleChangesLis
     private RuleWrapper ruleWrapper;
 
     private Composite simplePolygonComposite = null;
+
     private LinePropertiesEditor linePropertiesEditor;
+
     private Composite mainComposite;
+
     private StackLayout mainStackLayout;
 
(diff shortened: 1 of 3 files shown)


plugins/org.locationtech.udig.style.advanced.feature/src/org/locationtech/udig/style/advanced/lines/LinePropertiesComposite.java
```diff
@@ -417,6 +417,8 @@ public void onStyleChanged( Object source, String[] values, boolean fromField, S
                 } catch (Exception e) {
                     e.printStackTrace();
                 }
+            } else {
+            	ruleWrapper.getRule().setFilter(null);
```
sschulz92, line 421: Hey :) Can you please apply the code-formatter of uDig to all changed files?

sschulz92 (changes requested):

plugins/org.locationtech.udig.style.advanced.feature/src/org/locationtech/udig/style/advanced/lines/LinePropertiesComposite.java
AndreB87: Hey, shure - done ;)

fgdrf:
@AndreB87 Thanks for this fix and your first contribution to uDig!

I can confirm that it is an issue and the patch helps. It would be great to have a test-case as well, However, we do not have a test-bundle for style.advanced.feature. So I guess we can accept this fix since the class is not that easy to test.

PR #545, opened 2021-10-06
Apply code-formatter in module 'udig.tutorials.examples'

sschulz92:
This includes also:

- Some additions of "SuppressWarnings"
- Replacement of SubProgressMonitor with SubMonitor

PR #590, opened 2021-11-07
Use SubMonitor in udig.catalog.* modules and libs module

sschulz92:
The following parts are still missing and will be done in upcoming Pull-Requests:

1. udig.project
2. udig.project.ui
3. udig.tutorials.examples

In total there are 15 matches left :) We are getting closer!

### DIFF
--- a/plugins/org.locationtech.udig.catalog/src/org/locationtech/udig/catalog/IProcess.java
+++ b/plugins/org.locationtech.udig.catalog/src/org/locationtech/udig/catalog/IProcess.java
@@ -51,7 +51,7 @@ public abstract class IProcess implements IResolve {
     /**
      * Blocking operation to resolve into the adaptee, if available.
      * <p>
-     * Required adoptions:
+     * Required adaptions:
      * <ul>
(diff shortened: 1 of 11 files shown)


sschulz92:
The existing https://github.com/locationtech/udig-platform/pull/545 is handling the module `udig.tutorials.example` :)

plugins/org.locationtech.udig.catalog/src/org/locationtech/udig/catalog/IProcess.java
```diff
@@ -50,28 +51,30 @@
     /**
      * Blocking operation to resolve into the adaptee, if available.
      * <p>
-     * Required adaptions:
```
fgdrf, line 53: IMHO adaptions is correct here, I could not find adoptions in my dictionary ;)

fgdrf:
Looks good to me except from the functional perspective. Could you revert the adaptable spell-change and we are good to go

fgdrf (approved):

PR #592, opened 2021-11-08
Add missing 'break' to switch-case

sschulz92:


plugins/org.locationtech.udig.catalog/src/org/locationtech/udig/catalog/util/ASTFactory.java
```diff
@@ -314,48 +344,56 @@ public AST getRight() {
...
          * @see org.locationtech.udig.catalog.util.AST#getLeft()
          */
+        @Override
         public AST getLeft() {
             return null;
         }
-        /*
+
+        /**
          * (non-Javadoc)
```
fgdrf, line 387: IMHO we can remove "(non-Javadoc)" and "TODO summary sentence .." comments completely where `@Override` were added. Documentation should be in interfaces ..

@sschulz92 Are you kind to remove these with this pull?

plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/render/displayAdapter/impl/EventHandler.java
```diff
@@ -60,19 +66,22 @@ public void handleEvent( Event event ) {
             break;
         case SWT.Resize:
             controlResized(event);
+            break;
```
fgdrf, line 69: I assume we have no BUG here since other cases haven't implemented anything. So nice catch to avoid BUGS in the future!

plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/render/displayAdapter/impl/EventHandler.java
```diff
@@ -178,23 +188,23 @@ public void mouseMove( Event e ) {
     /**
      * @see org.eclipse.swt.events.MouseTrackListener#mouseEnter(org.eclipse.swt.events.MouseEvent)
      */
-    public void mouseEnter( Event e ) {
+    public void mouseEnter(Event e) {
```
fgdrf, line 191: Just wondering why `@Override` was added in other classes of this changes but not in this one. Please remove `@see` javadocs as well

fgdrf (changes requested):

plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/render/displayAdapter/impl/EventHandler.java
```diff
@@ -60,79 +60,73 @@ public void handleEvent( Event event ) {
             break;
         case SWT.Resize:
             controlResized(event);
+            break;
         case SWT.KeyDown:
-        // System.out.println("keydown");
+            break;
```
fgdrf, line 65: I guess since there is no action for key-down, this case-block for `SWT.KeyDown` should be removed

fgdrf (changes requested):
please remove "SWT.KeyDown" because of empty block

fgdrf (approved):

PR #593, opened 2021-11-08
Remove deprecated classes and use PanTool instead

sschulz92:


fgdrf:
@egouge can you please have a look at, since you added FixedScalePan in uDig 1.2. ..

Much appreciate your help!

fgdrf (approved):
Looks good to me! @egouge what do you think?

PR #595, opened 2021-11-08
Use 'Assert' of package 'org.eclipse.core.runtime'

sschulz92:


sschulz92:
@fgdrf do you mind having a look onto this PR? :)

fgdrf:
The changeset is fine, thanks for this. 

I guess the reason behind ist the deprecation of org.eclipse.jface.util.Assert

_As of 3.3, replaced by {@link org.eclipse.core.runtime.Assert}_

PR #597, opened 2021-11-09
Use ActionBars and remove deprecated method in IToolContext

sschulz92:


plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/tool/impl/ToolContextImpl.java
```diff
@@ -257,15 +198,7 @@ public RenderManager getRenderManagerInternal() {
...
     @Override
-    public IStatusLineManager getStatusBar() {
-        IActionBars2 bars = getActionBars();
-        if (bars == null)
-            return null;
-        return bars.getStatusLineManager();
-    }
-
-    @Override
-    public IActionBars2 getActionBars() {
```
fgdrf, line 268: Looks good to me to use IActionBars rather than IActionBars2 since CoolbarManager is never used in uDig.

PR #598, opened 2021-11-09
Enables advanced projection for features

sschulz92:


plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/render/RendererUtils.java
```diff
@@ -0,0 +1,51 @@
...
+        }
+        return Boolean.TRUE.equals(Boolean.valueOf(property.trim()));
+
+    }
+
+    /**
+     * @return true if renderer Jobs should be System-Jobs to avoid continuously showing 'Rendering
+     *         Map' in Progress View on streamed updates on Layer Data source.
+     */
+    public static boolean isRendererJobSystemJob() {
```
fgdrf, line 47: Looks like this changeset introduces two new options:
- advancedProjectionSupport, and
- set renderer jobs as system jobs ( to hide it in progress view)

out of curiosity is there a specific reason why the first is implemented as system property while the 2nd uses preferences store and options to configure Applications using `.options` files

plugins/org.locationtech.udig.render.feature.basic/src/org/locationtech/udig/render/internal/feature/basic/BasicFeatureRenderer.java
```diff
@@ -463,7 +471,6 @@ private void render(Graphics2D graphics, ReferencedEnvelope bounds, IProgressMon
             if (paintArea == null || paintArea.isEmpty() || validBounds == null
                     || validBounds.isEmpty() || validBounds.isNull() || validBounds.getWidth() <= 0
                     || validBounds.getHeight() <= 0) {
```
fgdrf, line 473: Maybe a trace option would be good here

plugins/org.locationtech.udig.render.feature.basic/src/org/locationtech/udig/render/internal/feature/basic/BasicFeatureRenderer.java
```diff
@@ -430,9 +430,17 @@ private void render(Graphics2D graphics, ReferencedEnvelope bounds, IProgressMon
             rendererHints.put(StreamingRenderer.LABEL_CACHE_KEY,
                     new LabelCacheDecorator(labelPainter, origin, layerId));
 
+            if (RendererUtils.isAdvancedProjectionSupportEnabled()) {
+                rendererHints.put(StreamingRenderer.ADVANCED_PROJECTION_HANDLING_KEY, true);
```
fgdrf, line 434: These option is for feature renderings such as shapefiles. What about raster datasets and WMS layers, any chance to allow enabling advanced projection support for these as well?

fgdrf (commented):
Thanks for this pull-request. Looks promising and I guess it’s worth to add some lines to release notes for 2.3.0 (what is new rst file)

fgdrf:
I'd like to split this one into two different pull requests. This will adress advanced projection support and the other one with allow to configure rendering jobs as system jobs. @sschulz92 would you agree?

sschulz92:
I do agree :) let's split it into two PR!

sschulz92:
@olsmn will create a new Pull Request with fixed issues.

PR #604, opened 2021-11-09
Fix possible NullPointerException in RenderManagerImpl

sschulz92:
Using another Renderer than uDig does per default, I come accross some possible NullPointerExceptions in RenderManagerImpl.

Also I added some more things to the `dispose()` call to clean-up even more.

Unfortunately I cannot provide any stacktrace.


plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/render/impl/RenderManagerImpl.java
```diff
@@ -740,15 +752,16 @@ public void enableRendering() {
 
     @Override
     public boolean isRenderingEnabled() {
-        return renderingEnabled;
+        return renderingEnabled && getMapDisplay() != null && getMapDisplay().getWidth() > 0
```
fgdrf, line 755: it would be great to provide a test-case as well ;)

fgdrf:
Thanks for collaborating!

PR #605, opened 2021-11-09
Fixes duplicated ModalTool in toolbar for ModalCategories

sschulz92:
In the application I am using, a MapView is implemented rather than a MapEditor. While developing stuff, I faced an issue that the toolbar of the view showed ModalTool's several times, which should not be the case.

Reproduction steps:

1. Add a shapefile as a layer
2. Create a new object using the edit-tools on this shapefile-layer
3. Close Map
4. Open the Map again (Show View / Map)
5. ModalTools are duplicated (ActionTools are not though!)
6. Select tool Zoom/Pan/Info --> no change, bug still remains
7. BUT: EditTool (Create Polygon) --> clean-up of toolbar is done, ModalTools are shown once

### DIFF
--- a/plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/tool/display/ModalToolCategory.java
+++ b/plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/tool/display/ModalToolCategory.java
@@ -1,7 +1,7 @@
-/*
- *    uDig - User Friendly Desktop Internet GIS client
- *    http://udig.refractions.net
- *    (C) 2004, Refractions Research Inc.
+/**
+ * uDig - User Friendly Desktop Internet GIS client
+ * http://udig.refractions.net
+ * (C) 2004, Refractions Research Inc.
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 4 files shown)


plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/tool/display/ModalToolCategory.java
```diff
@@ -84,61 +84,57 @@ protected IHandler getHandler() {
...
-                     * during maps opening/closing.
-                     */
-                    tool.removeContribution(old);
+    public void contribute(IToolBarManager manager) {
+        if (contribution == null) {
+            contribution = new CurrentModalToolContribution(getId());
+            for (ModalItem item : this) {
+                ToolProxy tool = (ToolProxy) item;
+                if (tool.getType() == ToolProxy.MODAL) {
+                    tool.addContribution(contribution);
```
fgdrf, line 115: Am I correct that addCorntribution ignores is if it has been added before. if so its not required to remove old first?

However, I was trying hard to reproduce the given workflow (edit layer, close map, re-open it, etc) but I couldn't. However it might be a problem if using a View rather than an editor. However, tested this fix and it doesn't break anything with MapEditorWithPalette - so good to go for me

fgdrf (approved):

PR #607, opened 2021-11-10
[HK] fix some issues in getting_started_Eclipse_IDE docu

olsmn:
Fixing of some issues noticed in the "Getting Started Eclipse IDE" documentation.


fgdrf:
Nice catch, Thanks for the fix!

PR #611, opened 2021-11-16
Feature to hide renderer job in progress views

fgdrf:
if the job is a system job, its not displayed in progress-view anymore for default. In Progress Views its possible to allow to show system jobs (and sleeping jobs) again

grafik

to configure it just use `.options` files it installation folder with 

`org.locationtech.udig.project/HIDE_RENDER_JOB=true`

is present.

Default behavior is as before : Jobs are still shown in Progress-View

### DIFF
--- a/docs/user/en/reference/Rendering Preferences.rst
+++ b/docs/user/en/reference/Rendering Preferences.rst
@@ -45,7 +45,23 @@ Available Preferences
    As stated this is an advanced option so do not modify it unless you understand the image formats
    and the ramifications of using one image type over another.
 
- 
+Other Preferences
+`````````````````
+
+.. _project_preferences-hide-renderer-job:
+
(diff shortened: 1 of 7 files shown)


olsmn:
Cool feature :)
Now one may choose to treat renderer Jobs as system jobs and so avoid the flickering caused by continuously showing "Rendering Map" in Progress View e.g. with streamed updates on Layer Data source.

olsmn:
However the first commit concerning pom.xml seems to be unrelated to the topic ...

fgdrf:
> However the first commit concerning pom.xml seems to be unrelated to the topic ...

you are right, seems I've mixed up changes from two different branches. I'll cleanup this pull-request and remove pom.xml from this changeset

olsmn:
Now that looks good :)

PR #612, opened 2021-11-16
Fix possible npe in render manager impl

fgdrf:
@olsmn & @sschulz92 this pull request adresses the initialization of Rendermanager in test-class.

since we do not have a Fragment the test-bundle only have access to API


fgdrf:
@olsmn @sschulz92 closing it since its a duplicate to #604 (with feedback that refactored test-case runs successfully) ;)

PR #613, opened 2021-11-18
Replace deprecated 'getImageData' with 'getImageData(int)'

sschulz92:
Within this PR I want to replace the deprecated method
```
@Deprecated
public ImageData getImageData() {
	return getImageData(100);
}
```
with the mentioned replacement:
```
getImageData(100);
```

This PR covers all projects, this deprecated method is gone afterwards.


fgdrf:
Looks like you have a run removing deprecation warnings. Thanks for your contribution.

This pull adresses deprecation warning and switches to `getImageData(100)`

Are yoiu kind to have a look at the test failure, is it related to the changeset?

fgdrf:
Created an issue for "flaky" test case #616

PR #615, opened 2021-11-19
remove Export-Package com.google.protobuf …

olsmn:
… from manifest of org.locationtech.udig.libs

Change-Id: I9597fcbd292f423dac457704729d5f65e2f174bd


fgdrf:
I guess its because the package is not used in udig bundles itself and might raise conflicts if uDig is integrated into other RCP application which has protobuf provided by another bundle.

Since the version is missing for the exported package the OSGi platform assumes the version is 2.3.0 (taking from bundle version) which is definitly wrong.

Thank you for thi sminor improvement!

PR #617, opened 2021-11-22
Use updated constructor of class FilteredTree

sschulz92:
Switch from deprecated constructor to new one:

from (deprecated):
```
@Deprecated
public FilteredTree(Composite parent, int treeStyle, PatternFilter filter) {
	super(parent, SWT.NONE);
	this.parent = parent;
	init(treeStyle, filter);
}
```

towards:
```
public FilteredTree(Composite parent, int treeStyle, PatternFilter filter, boolean useNewLook) {
	super(parent, SWT.NONE);
	this.parent = parent;
	this.useNewLook= useNewLook;
	init(treeStyle, filter);
}
```


fgdrf (commented):
In general the changeset looks good. 

Just a question : for "compatibility" reasons you used for useNewLook **false** which is fine. Can you provide screen-shots to see the difference between new and old look?

However, just for information: I stumbled about FilteredTree copies in codebase and filed an issue to review/fix :  #630

fgdrf:
I guess we can enable the new [look & feel](https://archive.eclipse.org/eclipse/downloads/drops/R-3.5-200906111540/eclipse-news-part3.html) later.

Thanks for the patchset!

PR #618, opened 2021-11-22
Replace deprecated method with newer method instead

sschulz92:
This PR will solve the deprecated use of `getAverageCharWidth()`.

Note that I needed to use `Math.round()` because the newer method is returning a `double` instead of an int.

Double to int conversion done like this:
```
(int) Math.round(gc.getFontMetrics().getAverageCharacterWidth())
```


fgdrf (changes requested):
Again, thanks for cleanup deprecations. Just a minor request for change regarding (unnecessary) cast to int.

fgdrf (approved):

PR #619, opened 2021-11-22
Remove method from interface ILayer and update all places accordingly

sschulz92:
Within this PR I removed `isType()` from interface ILayer.java which is deprecated. I updated all remaining places accordingly :)

"New" way to go: `hasResource()`

### DIFF
--- a/plugins/org.locationtech.udig.context/src/org/locationtech/udig/context/internal/ContextExportWizard.java
+++ b/plugins/org.locationtech.udig.context/src/org/locationtech/udig/context/internal/ContextExportWizard.java
@@ -91,9 +91,9 @@ public class ContextExportWizard extends Wizard implements IExportWizard {
         append(2, out, "<ResourceList>"); //$NON-NLS-1$
         for (ILayer layer : map.getMapLayers()) {
             try {
-                if (layer.isType(Layer.class)) {
+                if (layer.hasResource(Layer.class)) {
                     writeLayer(layer, out);
-                } else if (layer.isType(WFSDataStore.class)) {
+                } else if (layer.hasResource(WFSDataStore.class)) {
                     writeFeatureType(layer, out);
                 } else {
                     // n/a
--- a/plugins/org.locationtech.udig.project.edit/src/org/locationtech/udig/project/internal/provider/GlyphBuilder.java
+++ b/plugins/org.locationtech.udig.project.edit/src/org/locationtech/udig/project/internal/provider/GlyphBuilder.java
@@ -1,7 +1,7 @@
-/*
- *    uDig - User Friendly Desktop Internet GIS client
- *    http://udig.refractions.net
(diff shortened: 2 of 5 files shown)


fgdrf:
I had a look at the Implementations of ILayer and isType and hasResource were not implemented the same way:

WMSLaser:

https://github.com/locationtech/udig-platform/blob/4468dd6df7d6af80e8ef49b9bc7534d17f301452/plugins/org.locationtech.udig.info.tests/src/org/locationtech/udig/info/tests/WMSLayer.java#L229-L235

https://github.com/locationtech/udig-platform/blob/4468dd6df7d6af80e8ef49b9bc7534d17f301452/plugins/org.locationtech.udig.info.tests/src/org/locationtech/udig/info/tests/WMSLayer.java#L274-L277

whereas layerImpl is fine in my option:

https://github.com/locationtech/udig-platform/blob/bfd630ea0e8e435b034474263692daf169970847/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/impl/LayerImpl.java#L992-L1008

and LayerDecorator just delegates to the wrapped Layer which is fine to.

So I guess we at least WMSLayer should be adapted and I'd prefer to add a hint to release-notes for 2.3 as well (https://github.com/locationtech/udig-platform/blob/dca5ae79fdf5f90f343e80ef10c627e03b7cb076/docs/user/en/what_is_new/What%20is%20new%202.3.rst)

plugins/org.locationtech.udig.context/src/org/locationtech/udig/context/internal/ContextExportWizard.java
```diff
@@ -91,9 +91,9 @@ static void writeContext(IMap map, BufferedWriter out) throws IOException {
         append(2, out, "<ResourceList>"); //$NON-NLS-1$
         for (ILayer layer : map.getMapLayers()) {
             try {
-                if (layer.isType(Layer.class)) {
+                if (layer.hasResource(Layer.class)) {
                     writeLayer(layer, out);
```
fgdrf, line 95: wondering what would happen here, since WMSLayer.hasResource() is always false whereas isType was true ..

fgdrf (changes requested):
as stated in previous comment, at least WMSLayer might raise some issues because implementations of isType and hasResource aret different.

plugins/org.locationtech.udig.context/src/org/locationtech/udig/context/internal/ContextExportWizard.java
sschulz92: Basically WMSLayer.java is located into the module "org.locationtech.udig.info.tests" - so it's a test class. Does it really that matter? From my perspective we may even think about removing this class (should not belong to a test bundle).

PR #620, opened 2021-11-24
Move functions into MapPart

sschulz92:
… and add unimplemented method to related parts

In my project we use the `MapPart` instead of the `MapEditorPart` and we need those two methods. I guess there is nothing against moving this, @fgdrf what do you think?


plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/MapPart.java
```diff
@@ -76,4 +76,8 @@
      * @return Returns the DropHandler from this editor.
      */
     public UDIGDropHandler getDropHandler();
+
+    boolean isDragging();
```
fgdrf, line 80: Javadoc although we haven't any previously would be great ;)

plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/MapPart.java
```diff
@@ -76,4 +76,18 @@
...
+     * Returns true if drag and drop is enabled for the MapPart.
+     *
+     * @return Returns true if drag and drop is enabled for the MapPart.
+     */
+    boolean isDragging();
+
+    /**
+     * Enables or disables dragging (drag and drop) from the map editor.
+     *
+     * @param isDragging
```
fgdrf, line 90: javadoc for the parameter is missing ;)

fgdrf (approved):

PR #621, opened 2021-11-24
Add missing line in RenderExecutorComposite to set information

sschulz92:
if renderer job is system job or not


fgdrf:
Since its a subclass of renderJob it automatically uses setup from super implementation. With this changeset the same property would be set twice

would you agree?

PR #623, opened 2021-11-24
remove Export-Package org.geotools.data.geobuf …

olsmn:
…from manifest of org.locationtech.udig.libs

and move it as an optional extension to an own plugin org.locationtech.udig.libs.geobuf as fragment of org.locationtech.udig.libs

Change-Id: Ie3b67fe195d56fc0de8a4de00ccbed542177a312


fgdrf:
This its great, with this fragment the core module doesn't provide protobuf package anymore. This allows intergrators much easier to choose the prefered version of protobuf whereas the unsupported gt-geobuf is an optional feature.

Thanks for your effort making this possible!

PR #626, opened 2021-11-25
Use MapPart rather than MapEditorPart in ApplicationGIS/-Internal

sschulz92:


fgdrf:
tested locally with ArrowSelectionWithPopup and it still works as expected ✔️

PR #628, opened 2021-11-26
Rename MapEditorSite to MapSite to prevent misunderstandings

sschulz92:
Use MapSite in interface MapPart as in my project we are using the MapSite via this interface.

Within this PR I want to rename `MapEditorSite` to `MapSite` as I moved this implementation from `MapEditorPart` to `MapPart`. In my project the MapPart implementation was changed because a MapView is implemented.

@fgdrf what do you think?


plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/MapPart.java
```diff
@@ -70,6 +70,13 @@
      */
     IStatusLineManager getStatusLineManager();
 
+    /**
+     * Returns the MapSite of this editor.
```
fgdrf, line 74: returns the MapSite of the **MapPart** (no need to be an editor anymore ;))

plugins/org.locationtech.udig.tutorials.preferredzoomlevels/src/org/locationtech/udig/tutorials/preferredzoomlevels/View.java
```diff
@@ -69,6 +70,11 @@ public IStatusLineManager getStatusLineManager() {
         return getViewSite().getActionBars().getStatusLineManager();
     }
 
+    @Override
```
fgdrf, line 73: like in other tutorials, a mapSite instance should be created in createPartControl and used right here to access it:

```mapSite = new MapSite(getViewSite(), this);```

File: plugins/org.locationtech.udig.tutorials.tool-view/src/org/locationtech/udig/tutorials/toolview/View.java [hunk 3]
```diff
@@ -83,6 +84,11 @@ public IStatusLineManager getStatusLineManager() {
         return getViewSite().getActionBars().getStatusLineManager();
     }
 
+    @Override
+    public MapSite getMapSite() {
+        return null;
```
fgdrf, line 89: like in other tutorials, a mapSite instance should be created in createPartControl and used right here to access it:

mapSite = new MapSite(getViewSite(), this);

fgdrf (changes requested):

plugins/org.locationtech.udig.tutorials.preferredzoomlevels/src/org/locationtech/udig/tutorials/preferredzoomlevels/View.java
sschulz92: The whole getter is unneccessary as I had another look. This `View.java` extends the `DefaultMapViewPart` which is doing exactly what you are requesting. Therefore I removed this line here and in the other file.

plugins/org.locationtech.udig.tutorials.tool-view/src/org/locationtech/udig/tutorials/toolview/View.java
sschulz92: see above

fgdrf (approved):

PR #631, opened 2021-11-29
Update variable names and use MapPart rather than MapEditorPart

sschulz92:


plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/MapEditorPart.java
```diff
@@ -28,8 +28,6 @@
 
     Composite getComposite();
 
-    void setDirty(boolean b);
```
fgdrf, line 31: For now I‘d leave is/setDirty in MapEditorPart interface since it is editor related, isn‘t it.

fgdrf (changes requested):

plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/internal/MapEditorPart.java
sschulz92: A view can be dirty too, right?

sschulz92:
In our discussion we agreed to put the `setDirty()` method into the `MapPart`. Nevertheless we might review the whole implementation regarding this dirty-flag mode. The clean Eclipse way might look differently.

PR #633, opened 2021-11-29
Literal Strings should be compared using equals()

sschulz92:


fgdrf:
Thanks for this fix. Just had a look at apache (lang3) StringUtils which we depend on :

```
        if (bookmark.getName() == null || StringUtils.isBlank(bookmark.getName())) {
            bookmark.setName(Messages.BookmarkManager_bookmarkdefaultname + (++count));
        }
```

whould the check look like. What do you think, add another housekeeping commit for this?

PR #634, opened 2021-11-30
Remove MapEditor

sschulz92:
@fgdrf requested to remove this implementation from the repository as it seems that nobody is using this implementation anymore. To reduce the maintenance effort a little bit, I want to remove the `MapEditor.java`!

Additionally I renamed a method in interface `ViewportPane` to be aligned with recent changes and avoid misunderstandings:
from `MapPart getMapEditor()` towards `MapPart getMapPart()`.  With this change I needed to adapt all related locations also.


plugins/org.locationtech.udig.tool.select/src/org/locationtech/udig/tool/select/internal/AOILayerSelectionTool.java
```diff
@@ -235,16 +244,16 @@ public void run() {
         };
 
         @Override
-        protected IPreferenceStore fillFields( Composite parent ) {
+        protected IPreferenceStore fillFields(Composite parent) {
 
             Button nav = new Button(parent, SWT.CHECK);
-            nav.setText("Navigate");
+            nav.setText("Navigate"); //$NON-NLS-1$
```
fgdrf, line 250: this is an UI element and should be exported to allow translations, please remove `//$NON-NLS-1$` here ;)

plugins/org.locationtech.udig.tool.select/src/org/locationtech/udig/tool/select/internal/AOILayerSelectionTool.java
```diff
@@ -131,7 +136,8 @@ private Envelope getBounds( MapMouseEvent e ) {
     /**
      * @see org.locationtech.udig.project.ui.tool.SimpleTool#onMouseDragged(org.locationtech.udig.project.render.displayAdapter.MapMouseEvent)
      */
```
fgdrf, line 138: since you added `@Override` the javadoc with `@see` could be removed, could you adapt it as well?

fgdrf (changes requested):
Sometimes its hard to see the real change due to lots of reformatting. Because of the fact that the files are not all formatted with the formatter-setup it leads to changesets like this.

What I found and understood what this changeset is for:
* you removed mapEditor as requested on the mailing-list to avoid maintaining unused code
* you renamed ViewPortPane.getMapEditor to ViewPort.getMapPart

fgdrf:
I guess in addition to the code changes the sphinx developer docs need a bit love. Some RST files having code examples that would not be up-to-date anymore. Sections where MapEditor is used in a description its more a Concept rather than a concrete version we can leave at it is, IMHO.

sschulz92:
> I guess in addition to the code changes the sphinx developer docs need a bit love. Some RST files having code examples that would not be up-to-date anymore. Sections where MapEditor is used in a description its more a Concept rather than a concrete version we can leave at it is, IMHO.

I am not sure if I catched all places in the documentation. Do you mind have a second look please?

PR #637, opened 2021-12-02
correctly determine version in UiPlugin

olsmn:
in case the id of the product obtained from the Platform not equals UDIG_PRODUCT_ID 

Change-Id: I00ef55d38f743df73b513c8adb0f9bac2d1b281b


sschulz92 (approved):

fgdrf:
Thank you @olsmn for providing this changeset. Do I understand it correctly, it fixes an issue if uDig is not the product but udig used as SDK to integrate it into an existing RCP Application?

olsmn:
> Do I understand it correctly, it fixes an issue if uDig is not the product
> but udig used as SDK to integrate it into an existing RCP Application?
quite right :)

PR #639, opened 2021-12-07
Fix bug in if-condition

sschulz92:
The fix is hidden by the code-formatter, to clearify I changed the following part on purpose:

Old code:
```
AttributeType originalAttributeType = originalFeatureType.getType(name);
if (originalAttributeType == null
        && originalAttributeType instanceof GeometryType) {
    crs = ((GeometryType) originalAttributeType).getCoordinateReferenceSystem();
} else {
    crs = originalFeatureType.getCoordinateReferenceSystem();
}
```

New code:
```
AttributeType originalAttributeType = originalFeatureType.getType(name);
if (originalAttributeType != null
        && originalAttributeType instanceof GeometryType) {
    crs = ((GeometryType) originalAttributeType).getCoordinateReferenceSystem();
} else {
    crs = originalFeatureType.getCoordinateReferenceSystem();
}
```

Note the switch from `== null` towards `!= null`.

This will solve a SonarQube issue also, which detects this bug basically.


fgdrf:
Thanks a ton for this potential NullPointerException!

PR #642, opened 2021-12-10
Remove usage of IActionBars2 from MapSite

sschulz92:
Within uDig there is not other location where `IActionBars2` is implemented, the method `getCoolBarManager()` was not used once.

To be aligned with the whole project I would like to remove this `IActionBars2` implementation by using the `IActionBars` instead.


fgdrf:
tested with MapView implementation and it works, without this change I got strange ClassCastExceptions:

```
java.lang.ClassCastException: class org.eclipse.ui.internal.e4.compatibility.ActionBars cannot be cast to class org.eclipse.ui.IActionBars2 (org.eclipse.ui.internal.e4.compatibility.ActionBars and org.eclipse.ui.IActionBars2 are in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @18345249)
	at org.locationtech.udig.project.ui.internal.MapSite.getActionBars(MapSite.java:68)
	at org.locationtech.udig.project.ui.internal.tool.impl.ToolContextImpl.getActionBars(ToolContextImpl.java:206)
```

PR #643, opened 2021-12-10
Reactivate deactivated functions and add active ModalTool to MapViewer

sschulz92:
This re-activated functions are used in the context of the MapView implementation in the RCP standalone application I am involved.


plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/viewers/MapViewer.java
```diff
@@ -258,90 +256,81 @@ else if ( site instanceof IEditorSite){
...
-//        activeTool = tool;
-//        activeTool.setContext(getToolContext());
-//        activeTool.setActive(true); // this should register itself with the tool manager
-//        
-//        
-//        // this was normally handled by the ToolProxy which we cannot get a hold of
-//        String currentCursorID = activeTool.getCursorID();
-//		Cursor toolCursor = tools.findToolCursor(currentCursorID);
-//		getToolContext().getViewportPane().setCursor(toolCursor);
-//    }
```
fgdrf, line 307: I'm not sure about what you mean with reactivation functions. Looks like the whole message has a different behaviour than before. Can you explain why?

plugins/org.locationtech.udig.project.ui/src/org/locationtech/udig/project/ui/viewers/MapViewer.java
```diff
@@ -258,90 +256,81 @@ else if ( site instanceof IEditorSite){
...
-//        if( toolcontext == null ){
-//            toolcontext = new ToolContextImpl();
-//            toolcontext.setMapInternal(map);        
-//            toolcontext.setRenderManagerInternal(map.getRenderManagerInternal());            
-//        }
-//        return toolcontext;
-//    }
-    
-    public void setFont( Control control ) {
+    protected synchronized ToolContext getToolContext() {
```
fgdrf, line 299: ist used internally only, so lets keep it private

PR #644, opened 2021-12-10
Possibility to overwrite preferences using .options file

fgdrf:
This feature allows to overwrite preferences by filling .options as show below:

`plugin-id/preference-key=<new value>`

for example:

```org.locationtech.udig.project/HIDE_RENDER_JOB=true```

allows you to hide Render-Jobs shown in Progress-View (default is false).

Change-Id: Ib17c009e8ba0335af0bd9832dfa440c88a37954e

### DIFF
--- a/docs/user/en/tasks/Running uDig.rst
+++ b/docs/user/en/tasks/Running uDig.rst
@@ -82,6 +82,36 @@ You can change the location of your workspace using a command line option:
 
     udig -data <workspace location>
 
+Preferences
+-----------
+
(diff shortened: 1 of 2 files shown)


fgdrf:
@olsmn @sschulz92 I guess it would help to pre-configure uDig on a corporate level. I guess it might help if all possible preferences listed on preference user-doc page have at least the preference-key listed as well as the plugin-id where its hosted.

Great plus would be an example, especially for types such as colors and CRS's IDs.

What do you think?

sschulz92 (approved):

PR #646, opened 2021-12-12
Use 'equals' instead of '=='

sschulz92:
The remaining class is `MapViewer` which is handled in https://github.com/locationtech/udig-platform/pull/643 :)


plugins/org.locationtech.udig.catalog.worldimage/src/org/locationtech/udig/catalog/internal/worldimage/InMemoryCoverageLoader.java
```diff
@@ -115,36 +119,36 @@ private void updateMemoryLevel() throws IOException {
...
+        } else if (originalHeap < 1500) {
             heap = 2048;
         } else {
             heap = originalHeap * 2;
         }
-        
+
         String os = Platform.getOS();
-        if (heap > 1024 && os == Platform.OS_WIN32) {
+        if (heap > 1024 && os.equals(Platform.OS_WIN32)) {
```
fgdrf, line 129: Like it cursorPosition I'd suggest use `Platform.OS_WIN32.equals(Platform.getOS())` here

fgdrf (approved):

PR #648, opened 2021-12-14
Convert variable to be private to force users to use getter & setter

sschulz92:


plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/ProjectPlugin.java
```diff
@@ -123,25 +128,20 @@ public static Implementation getPlugin() {
...
         /**
          * Controls whether the warning message of non-undoable commands is shown.
-         *
-         * @deprecated Use the getter and setter methods.
          */
-        @Deprecated
-        public boolean undoableCommandWarning = true;
+        private boolean undoableCommandWarning = true;
 
         public void setUndoableCommandWarning(boolean value) {
```
fgdrf, line 146: Good to get rid of deprecation warnings here. However, since these getters/setters are public, a javadoc would be great.

As far as I remember there is MessageDialogWithToggle from eclipse ui to store decission kind of "show this dialog always or hide it in the future, because I'm aware of this". Not 100% sure about it but I guess this could made this setters/getters obsolete while the dialog stores it in workspace. Cool thing about it: after restart you not get anoyed anymore (again, again , and again :P)

Sounds more like a feature-request from my side 😄

fgdrf: here is an example how it is used within Eclipse IDE : https://git.eclipse.org/c/platform/eclipse.platform.ui.git/tree/bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEWorkbenchWindowAdvisor.java?id=e021405cf946127f11767e68968f484cb3ac3528

fgdrf (approved):

PR #436, opened 2020-12-09
F set scale cmd as nav command

fgdrf:
Cleanup and Refactor Code:
* removes deprecated Command Factories
* using packages w/o "internal" to allow to used Commands wherever required
* slitting up command implementation from API into seperate bundles, e.g. navigation, selection and edit commands

WORK IN PROGRESS!!

PR #649, opened 2021-12-14
Solve deprecation of ChangeCRSCommand

sschulz92:
Which ends in the deletion of BasicCommandFactory also


sschulz92:
I would like to ask for a review on this also @fgdrf :D

fgdrf:
This is related to #436 which I can update after this change.

so thanks for providing this cleanup of deprecated code. Wondering if its worth to add a hint in ‚What‘s new’ pages which Classes are not available anymore and what to use instead as a SDK user. Would you add a few lines?

I can have a look at other changes we have merged already

sschulz92:
> Wondering if its worth to add a hint in ‚What‘s new’ pages which Classes are not available anymore and what to use instead as a SDK user. Would you add a few lines?

I did, it's okay like this?

fgdrf (approved):
Great, thanks for your effort

fgdrf (approved):

PR #650, opened 2021-12-14
Remove several deprecated functions

sschulz92:
Within this PR I want to remove the following functions from the interface `ILayer` which are deprecated for quite a while:

1. `public boolean isSelectable();`
2. `void setSelectable(boolean value);`
3. `CoordinateReferenceSystem getCRS(IProgressMonitor monitor);`

### DIFF
--- a/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/Layer.java
+++ b/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/Layer.java
@@ -1,4 +1,5 @@
-/* uDig - User Friendly Desktop Internet GIS client
+/**
+ * uDig - User Friendly Desktop Internet GIS client
  * http://udig.refractions.net
  * (C) 2004-2012, Refractions Research Inc.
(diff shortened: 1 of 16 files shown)


plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/Layer.java
```diff
@@ -231,39 +231,6 @@
...
-     * layers are selectable, may be ignored for GeoResources that do not support editing. <!--
-     * end-user-doc -->
-     *
-     * @param value the new value of the '<em>Selectable</em>' attribute.
-     * @see #isSelectable()
-     * @deprecated use setInteraction(Interaction.SELECT, value)
-     */
-    @Deprecated
-    void setSelectable(boolean value);
-
```
fgdrf, line 266: these interface methods are generated from the model. I guess its necessary to remove these from EMF model and re-generate code usin EMF Toolset within Eclipse.

Maybe @egouge can support here?

PR #652, opened 2021-12-20
Use 'isEmpty()' rather than 'size() == 0'

sschulz92:
I want to replace `.size() == 0` calls with `isEmpty()`

Additionally I applied the code-formatter.


fgdrf (approved):
Thanks for this contribution, I assume these changes are suggestions by Sonarqube, aren't these?

PR #656, opened 2021-12-21
Trivial changes in package 'org.locationtech.udig.project.internal.impl'

sschulz92:
Following actions were made:
- Add `@Override` automatically
- Apply code-formatter
- Remove unnecessary javadoc
- Remove curly-braces for switch-case
- Fix possible NullPointerException in `ProjectImpl.java`

### DIFF
--- a/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/impl/MapImpl.java
+++ b/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/impl/MapImpl.java
@@ -1,4 +1,5 @@
-/* uDig - User Friendly Desktop Internet GIS client
+/**
+ * uDig - User Friendly Desktop Internet GIS client
  * http://udig.refractions.net
  * (C) 2004-2012, Refractions Research Inc.
  *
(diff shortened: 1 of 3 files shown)


plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/internal/impl/MapImpl.java
```diff
@@ -1407,36 +1431,35 @@ public void setRenderManagerInternalGen(RenderManager newRenderManagerInternal)
...
 
     Adapter adapter = new AdapterImpl() {
-        /**
-         * @see org.eclipse.emf.common.notify.impl.AdapterImpl#notifyChanged(org.eclipse.emf.common.notify.Notification)
-         */
         @Override
         public void notifyChanged(Notification msg) {
             switch (msg.getFeatureID(RenderManager.class)) {
-            case RenderPackage.RENDER_MANAGER__VIEWPORT_MODEL_INTERNAL: {
+            case RenderPackage.RENDER_MANAGER__VIEWPORT_MODEL_INTERNAL:
```
fgdrf, line 1437: IMHO the code looks much better but since it is a generated file I assume:
* its reset after generation from EMF model again
* its worth to find a way how the right formatter is applied after code-generation

I'd like to avoid efford the might be useless ..

sschulz92: Where can I find the information which class is generated? Is there any documentation I may read regarding EMF?

PR #658, opened 2021-12-22
Update all places to use 'isEmpty()'

sschulz92:
I replaced all missing modules to use `isEmpty()` instead of `.size() == 0`

Touched modules:
- `org.locationtech.udig.feature.editor`
- `org.locationtech.udig.project.ui`
- `org.locationtech.udig.project`

### DIFF
--- a/plugins/org.locationtech.udig.feature.editor/src/org/locationtech/udig/feature/panel/FeaturePanelRegistry.java
+++ b/plugins/org.locationtech.udig.feature.editor/src/org/locationtech/udig/feature/panel/FeaturePanelRegistry.java
@@ -1,4 +1,5 @@
-/* uDig - User Friendly Desktop Internet GIS client
+/**
+ * uDig - User Friendly Desktop Internet GIS client
  * http://udig.refractions.net
  * (C) 2010, Refractions Research Inc.
  * (C) 2001, 2008 IBM Corporation and others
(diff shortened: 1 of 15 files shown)


plugins/org.locationtech.udig.feature.editor/src/org/locationtech/udig/feature/panel/FeaturePanelRegistry.java
```diff
@@ -55,11 +53,11 @@
...
 public class FeaturePanelRegistry {
 
-    private static final String NO_TAB_ERROR = "Schema {0} declares non-existing tab {1}";
+    private static final String NO_TAB_ERROR = "Schema {0} declares non-existing tab {1}"; //$NON-NLS-1$
 
-    private static final String CONTRIBUTOR_ERROR = "Contributor {0} cannot be created.";
+    private static final String CONTRIBUTOR_ERROR = "Contributor {0} cannot be created."; //$NON-NLS-1$
 
-    private static final String TAB_ERROR = "Tab in {0} declares non-existing category {1}.";
+    private static final String TAB_ERROR = "Tab in {0} declares non-existing category {1}."; //$NON-NLS-1$
```
fgdrf, line 60: Looks like these constants are used to display messages in error view as IStatus (or at least in an error dialog). Therefore these should be translatable and might be exported to allow translations using different properties files

fgdrf (changes requested):
as stated, just remove the NLS comments right behind the three constants

fgdrf (approved):
Thanks a ton, ready to merge!

PR #660, opened 2021-12-28
Fix deprected usage if 'newInstance'

sschulz92:
Within the javadoc of `newInstance()` it is described how to fix the deprecation (Class.class):

```
clazz.newInstance() can be replaced by clazz.getDeclaredConstructor().newInstance()
```

This is what I did for all places where this deprecation occured.


sschulz92:
@fgdrf do you mind having a look onto this beauty? :)

PR #663, opened 2022-01-05
Remove unused imports from several Java classes

sschulz92:


sschulz92:
@fgdrf we might finish this topic of unused imports, what do you think? :) Some classes are still missing.

PR #670, opened 2022-01-09
Remove duplicated if-condition in 'FeatureInfoDisplay'

sschulz92:
The if-condition `value instanceof FeatureCollection` was used twice with the exact same content - detected by SonarQube :)


fgdrf (approved):
Nice catch! Has sonarqube guided you to this issue?

sschulz92:
> Nice catch! Has sonarqube guided you to this issue?

Basically I searched for bugs and code-smells with <3 occurrences so that I am able to remove them completely :D One of the bugs was, that the if-condition is totally the same so that the second if-condition will never be true.

fgdrf:
>  second if-condition will never be true.

I guess the compile might removed it, however it was dead code which never reached. many thanks again!

PR #671, opened 2022-01-10
[HK] removes unecessary Null-check

fgdrf:
@sschulz92 may you have a look at. fixes "Correctness - Nullcheck of value previously dereferenced" which means that the check is redundant.


sschulz92:
Nice 👍

PR #675, opened 2022-01-12
[HK][SONAR] using equals with hashCode

fgdrf:
Change-Id: I15c03a9a0668e55b9f20492e1f2769a53d1ee182

### DIFF
--- a/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/render/AbstractRenderMetrics.java
+++ b/plugins/org.locationtech.udig.project/src/org/locationtech/udig/project/render/AbstractRenderMetrics.java
@@ -12,6 +12,7 @@ package org.locationtech.udig.project.render;
 
 import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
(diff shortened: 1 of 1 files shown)


fgdrf:
this fixed sonar finding that classes that override equals() also override hashCode().

fgdrf:
@sschulz92 investigating if the bahavior of equals changed due to missing check of
```
        if (!(o instanceof AbstractRenderMetrics)) {
            return false;
        }
```
which has been replaced by 
```
        if (getClass() != obj.getClass()) {
            return false;
        }
```

I guess I will add a test-case to verify ..

sschulz92:
You have changed a missing check? I do not get the point tbh :D

fgdrf:
> You have changed a missing check? I do not get the point tbh :D

Eclipse allows to generate equals & hashCode using members of the class. the generated equals method change as mentioned and I'd like to check if tow different instances / sub-classes of this Abstract class returned equal=true in the past while the new version would say (different class) that these are not equals anymore.

HTH

Issue #678, opened 2022-02-02
NullPointerException in ShpGeoResourceInfor while executing search in SearchView


fgdrf:
What I did:
* add natruralearth dataset 10m-admin-0-countries.shp to map
* opened Search View and typed "countries" into search field
* exectued search, the following exception has been logged, nothing else happens:
```
!ENTRY org.locationtech.udig.catalog 4 0 2022-02-02 16:12:19.856
!MESSAGE Could not search in resource:C:\dev\data\testdata\10m-admin-0-countries.shp#10m-admin-0-countries
!STACK 0
java.lang.NullPointerException
	at java.net.URI$Parser.parse(URI.java:3042)
	at java.net.URI.<init>(URI.java:588)
	at org.locationtech.udig.catalog.internal.shp.ShpGeoResourceInfo.getSchema(ShpGeoResourceInfo.java:96)
	at org.locationtech.udig.catalog.internal.CatalogImpl.check(CatalogImpl.java:818)
	at org.locationtech.udig.catalog.internal.CatalogImpl.check(CatalogImpl.java:828)
	at org.locationtech.udig.catalog.internal.CatalogImpl.search(CatalogImpl.java:733)
	at org.locationtech.udig.catalog.ui.search.SearchView.searchImplementation(SearchView.java:513)
	at org.locationtech.udig.ui.SearchPart$1.run(SearchPart.java:530)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

Obviously doesn't have the the FeatureType a namespace set.

fgdrf:
After a debug-session it looks like that the ShpServiceExtension does't set ShapefileDataStoreFactory.NAMESPACEP.key in parameter map which leads to a Name-Object without a namespace.

Not sure why but the `ShpGeoResourceInfo.getSchema(ShpGeoResourceInfo.java:96)` does not handle null correctly, I guess I'll try to fix both ends.

PR #680, opened 2022-02-07
fixes NPE reported in issue 678

fgdrf:


sschulz92:
In the description you may refer to the issue to link them :)
https://github.com/locationtech/udig-platform/issues/678

plugins/org.locationtech.udig.catalog.db2/src/org/locationtech/udig/catalog/internal/db2/DB2GeoResource.java
```diff
@@ -199,7 +199,7 @@ protected DB2GeoResourceInfo createInfo(IProgressMonitor monitor) throws IOExcep
 
             icon = Glyph.icon(ft);
             keywords = new String[] { "db2", //$NON-NLS-1$
-                    ft.getName().getLocalPart(), ft.getName().getNamespaceURI() };
+                    ft.getName().getLocalPart(), ft.getName().getURI() };
```
fgdrf, line 202: should be similar to shp implementation. check for namespace, if not null, add it otherwise skip it

PR #683, opened 2022-07-05
HK : using sortpom for unique format pom.xml

fgdrf:
[Sortpom Maven Plugin](https://github.com/Ekryd/sortpom) helps to get and keep the same formation overall pom-files in the project. It sorts the xml elements dependent from the Maven model but is highly configurable in terms of what to format/sort and how.

Kudos to @aaime who [introduced](https://github.com/geotools/geotools/commit/ea2a21fa8633e73905e62eea9ac994cf5f455f76) the plugin in GeoTools project.

Whenever edited a pom file just run `mvn sortpom:sort -Dtycho.mode=maven -Dall -Pdocs,product,sdk` to format it before committing. Build verifies if the pom files need to be re-formated or already valid.


fgdrf:
@sschulz92 @olsmn What do you think?

fgdrf:
"Downgraded" from 3.x version to 2.15.0 which allows to build and verify with Java 1.8 as well

